PR #2, opened 2017-12-26
Add VPC Resource and Data Source

Savasw:
This pull request adds capability to work with VPC service in OpenTelekomCloud. It includes one resource, one datasource, unit tests, document, example and gophercloud sdk changes in vendor directory.


gator1:
Hi, right now the process is as follows:
All the checks first go to gator1/terraform-provider-opentelekomcloud. Only after some tests with customers we put the code here. 

gophercloud changes first go to gator1/gophercloud. I am in the process to separate gophercloud from Huawei/OTC specific changes. For now, the changes in gophercloud goes to gator1/gophercloud first. Then use govendor to put gator1 version of gophercloud into the gator1/ terraform-provider-opentelekomcloud. You need to provide a result of full acceptance test log that passed to be in, no exception.

gator1:
You don't check code into vendor, govendor is the tool to put the dependency code in the vendor directory. In gophercloud case, change gophercloud under gator1 first.

PR #5, opened 2018-03-01
delay to read resources

zengchen1024:
It should delay to read these resources, otherwise it may impact the performance of corresponding services.
This pr is from https://github.com/gator1/terraform-provider-opentelekomcloud/pull/15


zengchen1024:
@cgriggs01 Could you please merge this pr? Thanks very much!

paultyng:
I think we prefer using `resource.Retry` instead of explicit sleeps, it helps deal with usage inconsistencies for the provider. Basically you just either return a `resource.RetryableError` or `resource.NonRetryableError` from your func and it will retry until the timeout is reached.

zengchen1024:
@paultyng Thanks for your comments!  It sleep some time explicitly in order to delay to do subsequent operation. From my understanding,  **resource.Retry** should not be designed for this case. Am I right?

paultyng:
`Retry` is for the case where the API is potentially eventually consistent or the method returns before a resource is created (which I beleive is the same case here)? Here is an [example](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_internet_gateway.go#L52). It retries until the resource exists or the timeout is exhausted.

PR #9, opened 2018-03-02
add a new lib of golangsdk

zengchen1024:
this pr is for issue #8

PR #10, opened 2018-03-02
Create an instance of  ProviderClient defined in golangsdk

zengchen1024:
The new ProviderClient will be used to send/receive messages to opentelekom cloud

This pr is based on #9

### DIFF
--- a/opentelekomcloud/config.go
+++ b/opentelekomcloud/config.go
@@ -7,7 +7,6 @@ import (
 	"log"
 	"net/http"
 	"os"
-
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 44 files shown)


zengchen1024:
@cgriggs01 Could you please merge this pr? Thanks very much!

freesky-edward:
@paultyng @cgriggs01 could you please take a look at this PR. which proposes to add a new golangsdk client into config.go. the reason of adding this repository is to separate gophercloud as it was. so that it can have a better dependency management.

opentelekomcloud/config.go
```diff
@@ -198,6 +211,90 @@ func (c *Config) LoadAndValidate() error {
...
+		UserID:           c.UserID,
+	}
+
+	client, err := huaweisdk.NewClient(ao.IdentityEndpoint)
+	if err != nil {
+		return err
+	}
+
+	// Set UserAgent
+	client.UserAgent.Prepend(terraform.UserAgentString())
```
paultyng, line 207: Just FYI this method is moving to [`httpclient.UserAgentString()`](https://github.com/hashicorp/terraform/blob/master/httpclient/useragent.go) but its also worth mentioning that currently this only shares the vendored TF version, not the CLI version used to invoke the provider, although we are looking to address that.

opentelekomcloud/config.go
```diff
@@ -198,6 +211,90 @@ func (c *Config) LoadAndValidate() error {
...
+		caCertPool.AppendCertsFromPEM([]byte(caCert))
+		config.RootCAs = caCertPool
+	}
+
+	if c.Insecure {
+		config.InsecureSkipVerify = true
+	}
+
+	if c.ClientCertFile != "" && c.ClientKeyFile != "" {
+		clientCert, _, err := pathorcontents.Read(c.ClientCertFile)
```
paultyng: You may want to move some of this common initialization in to a shared func. There is a good amount of repetition between the two initialization functions. The azurerm provider [has an example](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/config.go) of how multiple client initialization has had a lot of the repetition cleaned up.

opentelekomcloud/config.go
```diff
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"os"
 
+	"strings"
```
paultyng: Got some unnecessary whitespace between `os` and `strings`

paultyng (changes requested):

opentelekomcloud/config.go
paultyng: I don't think you have to change this unless your vendored TF has the new method.

opentelekomcloud/config.go
zengchen1024: Done

opentelekomcloud/config.go
zengchen1024: Good suggestion! Thanks!

zengchen1024:
@paultyng Thanks very much for your comments! Could you take a look again. Thanks!

paultyng (approved):
This looks ok to me now, but could we add some acceptance testing? Or is it indirectly tested by other acceptance tests?

zengchen1024:
@paultyng This pr is tested indirectly by this pr #11 whose acceptance tests are all passed.

Issue #12, opened 2018-03-04
Error while creating instance


mvasilenko:
Hi there,

### Terraform Version
```
terraform version
Terraform v0.11.3
+ provider.opentelekomcloud v1.0.0
```
### Affected Resource(s)
Please list the resources as a list, for example:
- opentelekomcloud_compute_instance_v2

### Terraform Configuration Files
```hcl
provider "opentelekomcloud" {
  user_name   = "username"
  tenant_name = "eu-de"
  domain_name = "OTC-EU-DE-000000000010000XXXXX"
  password    = "password"
  auth_url    = "https://iam.eu-de.otc.t-systems.com/v3"
  region      = "eu-de"
}

# Create a web server
resource "opentelekomcloud_compute_instance_v2" "test-server" {
  name          = "test-server"
  image_name  = "Standard_CentOS_7_latest"
  flavor_name = "s1.medium"
}
```

### Debug Output

```
opentelekomcloud_compute_instance_v2.test-server: Still creating... (50s elapsed)
2018/03/04 13:54:44 [TRACE] dag/walk: vertex "meta.count-boundary (count boundary fixup)", waiting for: "opentelekomcloud_compute_instance_v2.test-server"
2018/03/04 13:54:44 [TRACE] dag/walk: vertex "provider.opentelekomcloud (close)", waiting for: "opentelekomcloud_compute_instance_v2.test-server"
2018/03/04 13:54:44 [TRACE] dag/walk: vertex "root", waiting for: "provider.opentelekomcloud (close)"
2018/03/04 13:54:49 [TRACE] dag/walk: vertex "meta.count-boundary (count boundary fixup)", waiting for: "opentelekomcloud_compute_instance_v2.test-server"
2018/03/04 13:54:49 [TRACE] dag/walk: vertex "provider.opentelekomcloud (close)", waiting for: "opentelekomcloud_compute_instance_v2.test-server"
2018/03/04 13:54:49 [TRACE] dag/walk: vertex "root", waiting for: "provider.opentelekomcloud (close)"
opentelekomcloud_compute_instance_v2.test-server: Still creating... (1m0s elapsed)
2018-03-04T13:54:52.023+0200 [DEBUG] plugin.terraform-provider-opentelekomcloud_v1.0.0_x4: 2018/03/04 13:54:52 [DEBUG] getAllInstanceNetworks: []opentelekomcloud.InstanceNetwork(nil)
[...]

PR #13, opened 2018-03-06
Add specific description for network parameter

freesky-edward:
The parameter of network in instance would have specific description. it should be a required parameter when there are multiple networks for the tenant. this PR proposes to update the description in doc.

Resolves: #12

### DIFF
--- a/website/docs/r/compute_instance_v2.html.markdown
+++ b/website/docs/r/compute_instance_v2.html.markdown
@@ -317,8 +317,9 @@ The following arguments are supported:
     the server. Changing this creates a new server.
 
 * `network` - (Optional) An array of one or more networks to attach to the
-    instance. The network object structure is documented below. Changing this
-    creates a new server.
+    instance. Required when there are multiple networks difined for the tenant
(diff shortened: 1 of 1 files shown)


website/docs/r/compute_instance_v2.html.markdown
```diff
@@ -317,8 +317,9 @@ The following arguments are supported:
     the server. Changing this creates a new server.
 
 * `network` - (Optional) An array of one or more networks to attach to the
-    instance. The network object structure is documented below. Changing this
-    creates a new server.
+    instance. Required when there are multiple networks difined for the tenant
```
paultyng: `Required when there are multiple networks defined for the tenant.` (fixed spelling on **defined** and added period at the end)

paultyng (changes requested):

website/docs/r/compute_instance_v2.html.markdown
freesky-edward: done

paultyng (approved):

PR #18, opened 2018-03-15
add common functions for building options

zengchen1024:
This pr refactors these methods at here[1][2][3],  The new methods will be more power in building req and resp parameters.

[1]https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/opentelekomcloud/ces_utils.go#L26
[2]https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/opentelekomcloud/ces_utils.go#L30
[3]https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/opentelekomcloud/ces_utils.go#L152

### DIFF
--- a/opentelekomcloud/build_param_util.go
+++ b/opentelekomcloud/build_param_util.go
@@ -1,12 +1,306 @@
 package opentelekomcloud
 
 import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"reflect"
+	"regexp"
+	"strings"
+
 	"github.com/hashicorp/terraform/helper/schema"
 )
+
+type funcSkipOpt func(optn string, jsonTags []string, tag reflect.StructTag) bool
+
+type exchangeParam struct {
+	OptNameMap *map[string]string
+	SkipOpt    funcSkipOpt
+}
+
+func (e *exchangeParam) toSchemaOptName(name string) string {
+	if e.OptNameMap != nil {
+		if n, ok := (*e.OptNameMap)[name]; ok {
+			return n
+		}
+	}
+	return strings.ToLower(name)
+}
+
+func (e *exchangeParam) BuildCUParam(opts interface{}, d *schema.ResourceData) ([]string, error) {
+	var skippedFields []string
+
+	optsValue := reflect.ValueOf(opts)
+	if optsValue.Kind() != reflect.Ptr {
+		return skippedFields, fmt.Errorf("parameter of opts should be a pointer")
+	}
+	optsValue = optsValue.Elem()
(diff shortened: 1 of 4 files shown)


paultyng:
Usage of reflection (`reflect` package) in Go is discouraged. Reflection is not clear, and I get why you would look to it for a provider to clean up some of this boilerplate, but that is not really how Go is written. Go favors verbosity over terse/clever code, that's why you write `if err != nil` over and over and over. It's not the way our providers are written, they manually should map all the fields, as it’s so different to the other providers - we’d be unable to accept a PR with reflection used to map the resource's schema.

Here are some thoughts from Rob Pike, one of the languages' creators, about reflections use cases (taken from his [Go Proverbs](https://go-proverbs.github.io/)): https://www.youtube.com/watch?v=PAAkCSZUG1c&t=14m35s

I'd be happy to discuss this further in the Slack room or over email.

paultyng:
One alternative we have seen people do is use code generation from reflection at build time to generate some of the mapping (the Google Cloud provider does this). So maybe that is a compromise you could pursue?

zengchen1024:
@paultyng  Thanks for your such constructive review!  Is the example of Google at https://github.com/terraform-providers/terraform-provider-google/blob/master/scripts/schemagen.go ?  I will try to write a similar tool.

paultyng:
I don't fully know how the Google provider generation works currently, but it may be related to it. They also have this OSS project: https://github.com/GoogleCloudPlatform/magic-modules to generate code for their SDK's and other projects.

Since Terraform typically requires more information than is present in Swagger/OpenAPI or other contracts or in SDK definitions, I think an ideal practice would be to generate your schema in a file, but allow it to be overridden manually when creating a resource. For example you do something like:

**resource_opentelekomcloud_compute_instance_v2.schema.go (generated)**
```golang
func resourceComputeInstanceV2Schema() *schema.Resource {
	return &schema.Resource{
		...
	}
}
```
**resource_opentelekomcloud_compute_instance_v2.go (manual)**
```golang
func resourceComputeInstanceV2() *schema.Resource {
	s := resourceComputeInstanceV2Schema()
	// modify schema here if necessary
	return s
}
```

This is just a suggestion though, not a requirement. But its very likely you will require a combination of generated and non-generated code, so whatever way you look in to should be able to accommodate that.

zengchen1024:
@paultyng You suggestions are very useful. My plan is to build a tool like Google at https://github.com/terraform-providers/terraform-provider-google/blob/master/scripts/schemagen.go. Developers can run this tool to generate an initial resource file if they want to save some time , then they can edit it to build a complete resource file.

PR #20, opened 2018-03-15
add an optional 'auto recovery' field to compute instance

zengchen1024:
TF_ACC=1 go test ./opentelekomcloud/ -v -run=TestAccComputeV2Instance_auto_recovery -timeout 240m
=== RUN   TestAccComputeV2Instance_auto_recovery
--- PASS: TestAccComputeV2Instance_auto_recovery (287.13s)
PASS
ok      github.com/terraform-providers/terraform-provider-opentelekomcloud/opentelekomcloud     287.139s

### DIFF
--- a/opentelekomcloud/resource_opentelekomcloud_compute_instance_v2.go
+++ b/opentelekomcloud/resource_opentelekomcloud_compute_instance_v2.go
@@ -346,6 +346,11 @@ func resourceComputeInstanceV2() *schema.Resource {
 				Type:     schema.TypeMap,
 				Computed: true,
 			},
+			"auto_recovery": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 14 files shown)


opentelekomcloud/resource_opentelekomcloud_compute_instance_v2.go
```diff
@@ -484,6 +489,14 @@ func resourceComputeInstanceV2Create(d *schema.ResourceData, meta interface{}) e
 		}
 	}
 
+	if hasFilledOpt(d, "auto_recovery") {
```
paultyng, line 492: Instead of creating the extra method, in Go you can just do an assignment and check inside the if:
```golang
if v, ok := d.GetOkExists("auto_recovery"); ok {
  ar := v.(bool)
  ...
}
```

paultyng (changes requested):
One minor comment about the wrapping function `hasFilledOpt`, but otherwise I think this is good.

opentelekomcloud/resource_opentelekomcloud_compute_instance_v2.go
zengchen1024: You are right under this scene which there is only one condition expression for 'if'. However if there are two or more condition expressions for 'if', such as 'if ** && ** ', then the wrapping method is  clearer. See the example[1]. So how about keeping it as it is?
[1] https://github.com/terraform-providers/terraform-provider-opentelekomcloud/pull/18/commits/aa77ebee4f8fc694a0d241d62d6b20aae5fe5bd9#diff-fa734f8e7b6638dfc32604a5e18322fdR259

paultyng: I admit that function preference is a subjective thing, so if you prefer it that is fine. In my opinion its not the typical way it would be approached in Go. I think Go very much prefers the verbose forms of things for clarity and repetition over terseness and dependency, etc. But I won't hold up the review on a preference.

**Edit:** I reposted the reflection comment on PR #18 for discussion there.

paultyng (approved):

PR #22, opened 2018-03-16
revise CES-alarm rule

zengchen1024:
This pr wants to use the lib of golangsdk instead of gophercloud. Actually, the official gophercloud does not include the lib of ces. Besides it removes the original helper methods which are used  in CRUD operation to build parameters, becuase they used reflect module of go,  which are not considered as a good way.

TF_ACC=1 go test ./opentelekomcloud/ -v -run=TestCESAlarmRule_basic -timeout 240m
=== RUN   TestCESAlarmRule_basic
--- PASS: TestCESAlarmRule_basic (148.84s)
PASS
ok      github.com/terraform-providers/terraform-provider-opentelekomcloud/opentelekomcloud     148.849s


zengchen1024:
@paultyng Could you take a look at this pr. Thanks very much!

opentelekomcloud/resource_opentelekomcloud_ces_alarmrule.go
```diff
@@ -334,17 +334,75 @@ func resourceAlarmRule() *schema.Resource {
 	}
 }
 
+func getMetricOpts(d *schema.ResourceData) alarmrule.MetricOpts {
+	mos := d.Get("metric").([]interface{})
```
paultyng: This is probably fine, but may be brittle or cause panics if you get unexpected `nil`s, etc. As long as you are comfortable that you don't need to use the type assertions with the `ok` its fine with me, but just something to be sure about, as panics are obviously very bad for user's state.

opentelekomcloud/utils.go
```diff
@@ -18,3 +22,36 @@ func jsonBytesEqual(b1, b2 []byte) bool {
...
+// convertStructToMap converts an instance of struct to a map object, and
+// changes each key of fileds to the value of 'nameMap' if the key in it
+// or to its corresponding lowercase.
+func convertStructToMap(obj interface{}, nameMap map[string]string) (map[string]interface{}, error) {
+	b, err := json.Marshal(obj)
+	if err != nil {
+		return nil, fmt.Errorf("refreshResourceData:: marshal failed:%v", err)
+	}
+
+	m := regexp.MustCompile(`"[a-z0-9A-Z_]+":`)
```
paultyng: move this outside the `func` so it can fail earlier if a problem exists

paultyng (changes requested):
One minor change for the `regexp` and one comment on type assertions, but overall looks ok to me.

opentelekomcloud/utils.go
zengchen1024: Sorry, I am not quite understand. Do you mean move line 35 outside of method of 'convertStructToMap'? Why? Line 35 will not be panic because it passes an raw string to 'regexp.MustCompile'.

opentelekomcloud/resource_opentelekomcloud_ces_alarmrule.go
zengchen1024: Done

opentelekomcloud/utils.go
paultyng: It can panic if the string is a bad regex. See [the docs](https://golang.org/pkg/regexp/#MustCompile), the `Must*` methods are typically for global initialization:

> MustCompile is like Compile but panics if the expression cannot be parsed. It simplifies safe initialization of global variables holding compiled regular expressions.

zengchen1024: It may panic if regexp.MustCompile receives a variable whose value is inconclusive. But at line 35, it receives a valid regular express, so it can not panic. I can use regexp.Compile instead of it, which will not lead to panic.

paultyng (approved):
This is fine like this, I maybe wasn't very clear in my comments about the regex compilation. I understand your string now is valid and won't panic, but in the future if someone changes it, its possible the panic can be missed until runtime depending on test coverage. If you have a constant regexp, the standard practice in Go is to use `MustCompile` but as a global var in your package. If you have something that could panic, especially if it changes, it better for user's of the provider if that panic happens in testing, and global vars are more likely to be run than every function. Compilation of the regexp isn't extremely expensive but it is [non-trival](https://golang.org/src/regexp/syntax/parse.go?s=19236:19286#L691) so it is worth only doing it once if you can.

Anyway, I think this is fine now as its resistant to panic / bad regex changes down the road as well, so will merge it.

PR #24, opened 2018-03-22
Add kms for volume in doc

zhongjun2:


website/docs/r/blockstorage_volume_v2.html.markdown
```diff
@@ -46,6 +51,14 @@ The following arguments are supported:
 
 * `metadata` - (Optional) Metadata key/value pairs to associate with the volume.
     Changing this updates the existing volume metadata.
+    The key/value include the following parameters:
+    `__system__cmkid` - (Optional) The ID of the kms key.
+
+    `region` - (Optional) The region of the kms key.
+
+    '__system__encrypted' - The default value is set to '0', it means
```
niuzhenguo, line 59: should be `` instead of '' here like above, and maybe need to add a (Optional) as well.

PR #29, opened 2018-03-26
fix the format error of log

zengchen1024:


opentelekomcloud/resource_opentelekomcloud_dns_zone_v2.go
```diff
@@ -289,7 +289,7 @@ func resourceDNSZoneV2ValidType(v interface{}, k string) (ws []string, errors []
 }
 
 func parseStatus(rawStatus string) string {
-	log.Printf("[DEBUG] OpenTelekomCloud DNS Zone (%s) raw status: %s", rawStatus)
+	log.Printf("[DEBUG] OpenTelekomCloud DNS Zone raw status: %s", rawStatus)
```
niuzhenguo: Is this redundant with L309? seems you only call parseStatus after log the same thing at L309. Or maybe you intend to log rawStatus and parsed status separately?

zengchen1024: yes, line 292 should be deleted.

niuzhenguo: Thanks for the update, LGTM now!

paultyng (approved):

PR #43, opened 2018-04-09
fix the bug of creating vm failed

zengchen1024:
for issue #44
The reason is that the service of auto recovery is not online at some region. To keep backwards compatibility, it should check the error type when reading auto recovery failed.

=== RUN   TestAccComputeV2Instance_auto_recovery
--- PASS: TestAccComputeV2Instance_auto_recovery (222.13s)
PASS
ok      github.com/terraform-providers/terraform-provider-opentelekomcloud/opentelekomcloud     222.136s


niuzhenguo:
LGTM!

zengchen1024:
@niuzhenguo Could you take at this pr? Thanks!

PR #50, opened 2018-04-11
Use golang for blockstorage tags

zengchen1024:
=== RUN   TestAccBlockStorageV2Volume_tags
--- PASS: TestAccBlockStorageV2Volume_tags (114.58s)


zengchen1024:
@niuzhenguo Could you take a look at this pr? Thanks!

opentelekomcloud/golangsdk_utils.go
```diff
@@ -12,3 +12,15 @@ func chooseECSV1Client(d *schema.ResourceData, config *Config) (*golangsdk.Servi
 func chooseCESClient(d *schema.ResourceData, config *Config) (*golangsdk.ServiceClient, error) {
 	return config.loadCESClient(GetRegion(d, config))
 }
+
+func chooseEVSV2Client(d *schema.ResourceData, config *Config) (*golangsdk.ServiceClient, error) {
```
niuzhenguo: not sure why we need to wrap the client with a "chooseclient" but actually not do anything expect returning the config.client directly. How about just using the config.loadEvSv2Client directly like other clients?

zengchen1024: There is no special reason. Personally, I like clear code. If invoking config.loadCESClient directly, it needs write 'GetRegion(d, config)'  every time.

zengchen1024: There is no special reason. Personally, I like clear code. If invoking config.loadCESClient directly, it needs write 'GetRegion(d, config)'  every time.

niuzhenguo: if that's the only reason, you could combine this with config.loadCESClient() instead of declaring two functions. and we should better to keep same format for all clients.

zengchen1024: sorry, not quite understand. how to combine?

niuzhenguo: something like this:
func (c *Config) loadEVSV2Client(d* schema.ResourceData) (*golangsdk.ServiceClient, error) {
        region = GetRegion(d, c)
	return huaweisdk.NewBlockStorageV2(c.HwClient, golangsdk.EndpointOpts{
		Region:       c.determineRegion(region),
		Availability: c.getHwEndpointType(),
	})
}

zengchen1024: Done

PR #98, opened 2018-07-17
Add DeH service

frisky-dis:
This PR adds a capability to manage OTC's DeH service.


opentelekomcloud/provider_test.go
```diff
@@ -30,6 +30,8 @@ var (
 	OS_VPC_ID                 = os.Getenv("OS_VPC_ID")
 	OS_SUBNET_ID              = os.Getenv("OS_SUBNET_ID")
 	OS_TENANT_ID              = os.Getenv("OS_TENANT_ID")
+	OS_SERVER_ID              = os.Getenv("OS_SERVER_ID")
+	OS_DEH_HOST_ID            = os.Getenv("OS_DEH_HOST_ID")
```
niuzhenguo: Please create a DEH resource with the test instead of adding a new ENV variable.

niuzhenguo: And the OS_SERVER_ID

Savasw: @niuzhenguo We couldn't find a way to associate ecs server with DeH using existing server resource as it doesn't support dedicated_host_id param which API supports. Hence creating it in acc test, won't be useful for actual test. We can skip test than failing it if that works for you. What do you suggest?

niuzhenguo: @Savasw I'm ok with skipping the tests for now! But as you said we can't associate ecs servers with DeH on terraform side, how do terraform users work with that?

niuzhenguo: @Savasw I just checked the API docs https://docs.otc.t-systems.com/en-us/api/ecs/en-us_topic_0068473331.html#EN-US_TOPIC_0068473331__en-us_topic_0057972661_table12534817105641 . We support to specify the dedicated_host_id with scheduler_hints param when creating instances, so I think you should also need to add it support here https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/opentelekomcloud/resource_opentelekomcloud_compute_instance_v2.go#L269

niuzhenguo:
@disha-wani Please resolve conflicts, thanks!

PR #100, opened 2018-07-24
Some parameters should be computed as they are optional

mcanevet:


niuzhenguo:
@mcanevet why does parameters should be computed as they are optional? Value of the parameters can either comes from the config, is computed, or is both.

mcanevet:
If I don't set it it is not idempotent if it not computed.

niuzhenguo:
@zengchen1024 can you please help to check such optional ELB parameters?

PR #101, opened 2018-07-24
Add BMS service

simrannagra:
This PR adds a capability to manage OTC's BMS service.

### DIFF
--- a/opentelekomcloud/data_source_opentelekomcloud_compute_bms_nic_v2_test.go
+++ b/opentelekomcloud/data_source_opentelekomcloud_compute_bms_nic_v2_test.go
@@ -40,8 +40,33 @@ func testAccCheckBMSNicV2DataSourceID(n string) resource.TestCheckFunc {
 }
 
 var testAccOpenTelekomCloudBMSNicV2DataSource_basic = fmt.Sprintf(`
+resource "opentelekomcloud_vpc_v1" "vpc_1" {
+  name = "vpc_otc123"
+  cidr = "192.168.0.0/16"
+}
+
+resource "opentelekomcloud_vpc_subnet_v1" "subnet_1" {
+  name = "sub_otc123"
+  cidr = "192.168.0.0/16"
+  gateway_ip = "192.168.0.1"
+  vpc_id = "${opentelekomcloud_vpc_v1.vpc_1.id}"
+  availability_zone = "%s"
+}
+resource "opentelekomcloud_compute_instance_v2" "instance_1" {
+  name = "BMSinstance_1"
+  image_id = "%s"
+  security_groups = ["default"]
+  availability_zone = "%s"
+  flavor_id = "physical.o2.medium"
+  metadata {
+    foo = "bar"
+  }
+  network {
+    uuid = "${opentelekomcloud_vpc_subnet_v1.subnet_1.id}"
+  }
+}
 data "opentelekomcloud_compute_bms_nic_v2" "nic_1" {
-  server_id = "%s"
+  server_id = "${opentelekomcloud_compute_instance_v2.instance_1.id}"
   id = "%s"
 }
-`, OS_SERVER_ID, OS_NIC_ID)
+`, OS_AVAILABILITY_ZONE, OS_IMAGE_ID, OS_AVAILABILITY_ZONE, OS_NIC_ID)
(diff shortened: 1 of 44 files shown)


opentelekomcloud/provider_test.go
```diff
@@ -30,6 +30,10 @@ var (
 	OS_VPC_ID                 = os.Getenv("OS_VPC_ID")
 	OS_SUBNET_ID              = os.Getenv("OS_SUBNET_ID")
 	OS_TENANT_ID              = os.Getenv("OS_TENANT_ID")
+	OS_KEYPAIR_NAME           = os.Getenv("OS_KEYPAIR_NAME")
+	OS_BMS_FLAVOR_NAME        = os.Getenv("OS_BMS_FLAVOR_NAME")
+	OS_SERVER_ID              = os.Getenv("OS_SERVER_ID")
```
niuzhenguo: Instead of adding a new OS_SERVER_ID, can you please creating a new bms server with the test case?

opentelekomcloud/provider_test.go
```diff
@@ -316,3 +331,24 @@ func envVarFile(varName string) (string, error) {
 	}
 	return tmpFile.Name(), nil
 }
+
+func testAccBmsKeyPairPreCheck(t *testing.T) {
+	testAccPreCheckRequiredEnvVars(t)
+	if OS_KEYPAIR_NAME == "" {
+		t.Fatalf("Provide the key pair name")
```
niuzhenguo: Can we just skip the keypair data source tests if it's not set?

opentelekomcloud/provider_test.go
```diff
@@ -316,3 +331,24 @@ func envVarFile(varName string) (string, error) {
...
+	}
+}
+
+func testAccBmsFlavorPreCheck(t *testing.T) {
+	testAccPreCheckRequiredEnvVars(t)
+	if OS_BMS_FLAVOR_NAME == "" {
+		t.Fatalf("Provide the bms name starting with 'physical'")
```
niuzhenguo: ditto.

opentelekomcloud/provider_test.go
```diff
@@ -141,6 +145,17 @@ func testAccPreCheckSwift(t *testing.T) {
...
 }
 
+func testAccPreCheckBMSNic(t *testing.T) {
+	testAccPreCheckRequiredEnvVars(t)
+
+	if OS_SERVER_ID == "" {
+		t.Fatal("OS_SERVER_ID must be set for NIC acceptance tests")
+	}
+	if OS_NIC_ID == "" {
+		t.Fatal("OS_NIC_ID must be set for NIC acceptance tests")
```
niuzhenguo: Can we just skip the data source tests if this is not set?

opentelekomcloud/data_source_opentelekomcloud_compute_bms_nic_v2_test.go
```diff
@@ -0,0 +1,72 @@
...
+  name = "BMSinstance_1"
+  image_id = "%s"
+  security_groups = ["default"]
+  availability_zone = "%s"
+  flavor_id = "physical.o2.medium"
+  metadata {
+    foo = "bar"
+  }
+  network {
+    uuid = "${opentelekomcloud_vpc_subnet_v1.subnet_1.id}"
```
niuzhenguo: You can use OS_NETWORK_ID directly for testing instead of creating vpc, subnet resource within every test case.

niuzhenguo (commented):
Besides the comments inline, I always got an error like "After applying this step, the plan was not empty". I wonder if you have ever successfully passed the tests?

Savasw:
@niuzhenguo we were not able to reproduce the 'Plan was not empty error', however there indeed was one logical issue with the nic test after our last commit to add vpc and subnet, which got missed by us. We have tested all the bms acc tests, and they all are passing at our end. Can you please try running them again at your end and let us know if you still are facing that issue. I understand it takes awfully lot of time to run bms test due to its creation time, we regret the inconvenience caused.

niuzhenguo:
@Savasw The reason why I got 'Plan was not empty error' is because 'OS_FLAVOR_NAME' is specified on my test env, and then resource_compute_instance uses it as a default value for flavor_name param if not specified https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/opentelekomcloud/resource_opentelekomcloud_compute_instance_v2.go#L78 . So to make tests happy for all enviroments, can you please specify both flavor_id and flavor_name with your test case? Sorry for the inconvenience!

Issue #105, opened 2018-08-27
Adding/Removing Secgroup+rule forces new resource


Moep90:
### Terraform Version
```
$ terraform -v
Terraform v0.11.7
```

### Affected Resource(s)
- opentelekomcloud_compute_instance_v2
- opentelekomcloud_networking_secgroup_rule_v2
- opentelekomcloud_networking_secgroup_rule_v2

### Expected Behavior
Adding or removing a Secgroup, Secrule and applying them without replacing instance

### Actual Behavior
Terraform wants to create a new resource.. With new block-device etc etc...

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
# Instance defenition
```HCL
resource "opentelekomcloud_compute_instance_v2" "<project>_instances_az0" {
  count             = "${var.<project>_instances_az0_count}"
  name              = "${var.project}-${terraform.workspace}-${var.availability_zones[0]}-node${format("%02d", count.index+1)}"
  image_name        = "${var.image}"
  flavor_name       = "${var.flavor_name}"
  key_pair          = "${opentelekomcloud_compute_keypair_v2.keypair.name}"
  security_groups   = ["${var.project}-${terraform.workspace}-<secgroup3>",
                       "${var.project}-${terraform.workspace}-<secgroup4>"]
  region            = "${var.region}"
  availability_zone = "${var.availability_zones[0]}"

  metadata {
    this = "Projekt: ${var.project} Stage: ${terraform.workspace}"
  }

  block_device {
    uuid              = "${data.opentelekomcloud_images_image_v2.<osname>.id}"
    source_type           = "image"
    volume_size           = "${var.<project>_systemdisk_size}"
    boot_index            = 0
    destination_type      = "volume"
    delete_on_termination = false
  }

  network {
    uuid        = "${opentelekomcloud_networking_network_v2.vpc.id}"
    fixed_ip_v4 = "${var.<project>_instance_priv_ips_az0}${count.index + 1}"
  }
}
```
# Output from plan
```HCL
-/+ opentelekomcloud_compute_instance_v2.<project-name>_instances_az0[1] (new resource required)
      id:                                   "388<id>27d" => <computed> (forces new resource)
[...]
      availability_zone:                    "eu-de-01" => "eu-de-01"
      block_device.#:                       "1" => "1"
      block_device.0.boot_index:            "0" => "0"
      block_device.0.delete_on_termination: "false" => "false"
[...]

PR #186, opened 2019-01-11
Fix couple of docs issue

niuzhenguo:
1. Remove 'tags' from s3_bucket lifecycle_rule doc
2. Mark security_group as non-updatable on compute_instance doc


niuzhenguo:
#105

PR #232, opened 2019-02-25
Add agency resource

zhongjun2:


opentelekomcloud/resource_opentelekomcloud_identity_agency_v3.go
```diff
@@ -0,0 +1,592 @@
...
+		},
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(10 * time.Minute),
+			Update: schema.DefaultTimeout(10 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": {
```
zengchen1024: it is better to remove the region, or not set it as 'ForceNew', because the region parameter is not used any more

opentelekomcloud/resource_opentelekomcloud_identity_agency_v3.go
```diff
@@ -0,0 +1,592 @@
...
+				ForceNew: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"delegated_domain_name": {
+				Type:     schema.TypeString,
+				Required: true,
```
zengchen1024, line 44: miss 'ForceNew: true,'

opentelekomcloud/resource_opentelekomcloud_identity_project_v3.go
```diff
@@ -47,6 +47,16 @@ func resourceIdentityProjectV3() *schema.Resource {
 				Computed: true,
 				ForceNew: true,
 			},
+
```
zengchen1024, line 50: why add enabled, and is_domain as you delete the check assert of them

opentelekomcloud/provider_test.go
```diff
@@ -35,6 +35,7 @@ var (
 	OS_BMS_FLAVOR_NAME        = os.Getenv("OS_BMS_FLAVOR_NAME")
 	OS_NIC_ID                 = os.Getenv("OS_NIC_ID")
 	OS_TO_TENANT_ID           = os.Getenv("OS_TO_TENANT_ID")
+	OS_TENANT_NAME            = os.Getenv("OS_TENANT_NAME")
```
zengchen1024: OS_TENANT_NAME may be fetch from environment of "OS_PROJECT_NAME". see  https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/opentelekomcloud/provider.go#L71-L73

opentelekomcloud/resource_opentelekomcloud_identity_agency_v3.go
zhongjun2: The delegated_domain_name will be updated. so we don't need 'ForceNew: true,'

opentelekomcloud/resource_opentelekomcloud_identity_agency_v3.go
zhongjun2: Delete region

opentelekomcloud/resource_opentelekomcloud_identity_project_v3.go
zhongjun2: Because it will show in project resource response after create a project

opentelekomcloud/provider_test.go
zhongjun2: done

Issue #257, opened 2019-04-01
[RDS] HA-Flavor unable to use two AZs


Moep90:
### Terraform Version
```hcl
$ terraform --version
Terraform v0.11.13
+ provider.dns v2.0.0
+ provider.local v1.2.0
+ provider.null v2.1.0
+ provider.openstack v1.17.0
+ provider.opentelekomcloud v1.7.0
+ provider.template v2.1.0
```
### Affected Resource(s)
- opentelekomcloud_rds_instance_v1

### Terraform Configuration Files
```hcl
data "opentelekomcloud_rds_flavors_v1" "rds_flavor" {
    region            = "${var.region}"
    datastore_name    = "MySQL"
    datastore_version = "${var.rds_datastore_version}"
    speccode          = "rds.mysql.s1.medium.ha"
}

resource "opentelekomcloud_rds_instance_v1" "instance" {
  name              = "${var.project}-${terraform.workspace}-rds-instance"
  region            = "${var.region}"
  vpc               = "${opentelekomcloud_networking_router_v2.router.id}"
  availabilityzone  = "${var.availability_zones[0]}"

  dbport            = "${var.rds_db_port}"
  dbrtpd            = "${var.rds_db_root_user_pw}"

  backupstrategy = {
    starttime       = "04:00:00"
    keepdays        = "${var.rds_backup_keepdays}"
  }

  ha = {
    enable          = true
    replicationmode = "async"
  }

  datastore {
    type            = "MySQL"
    version         = "${var.rds_datastore_version}"
  }
  flavorref         = "${data.opentelekomcloud_rds_flavors_v1.rds_flavor.id}"

  volume {
    type            = "COMMON"
    size            = "${var.rds_db_size}"
  }
  nics {
    subnetid        = "${opentelekomcloud_networking_network_v2.vpc.id}"
  }

  securitygroup {
    id              = "${opentelekomcloud_networking_secgroup_v2.inside_rds_traffic.id}"
  }

  lifecycle {
   prevent_destroy = "true"
   ignore_changes  = ["name"]
  }
}
```

### Expected Behavior
Terraform provides the first instance of the HA-RDS in the givven AZ and adds the standby to another AZ.

### Actual Behavior
Terraform provides both HA-RDS (Prime/Standby) into the same AZ.

### Steps to Reproduce
1. `terraform apply`

niuzhenguo:
[...]

Issue #251, opened 2019-03-26
[REQUEST] Support for `disk_encryption_id` implemented in RDS API v3


tstrohmeier:
Hi!

Last week on the 19th March, OTC v4 was released and brought us a new RDS API v3. We would like to see a support for the API v3, in this provider, since we need the newly exposed property `disk_encryption_id` very urgent. 

Here is the link to API v3: https://docs.otc.t-systems.com/en-us/api/rds/en-us_topic_0160329112.html

My proposal on how to extend the RDS resource:
```hcl
resource "opentelekomcloud_kms_key_v1" "key_1" {
  key_alias       = "key_1"
  realm           = "eu-de"
}

resource "opentelekomcloud_rds_instance_v1" "pg_instance" {
  name = "rds-xyz"
  ...
  disk_encryption_id = "${opentelekomcloud_kms_key_v1.key_1.id}"
  ...
  depends_on = ["opentelekomcloud_kms_key_v1.key_1"]
}
```

Thank you,
Thomas

PR #267, opened 2019-04-25
Rds v3

zengchen1024:
--- PASS: TestAccRdsInstanceV3_basic (544.25s)
--- PASS: TestAccOpenTelekomCloudRdsFlavorV3DataSource_basic (100.86s)
--- PASS: TestAccRdsInstanceV3_importBasic (495.29s)


opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
```diff
@@ -0,0 +1,1004 @@
...
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"availability_zone": {
```
tstrohmeier, line 39: If I understand the docs correct the following format is allowed? 
`availability_zone = "{{ availability_zone_1 }},{{ availability_zone_2 }}"`

I think it would be better to use a array:
`availability_zone = ["{{ availability_zone_1 }}" , "{{ availability_zone_2 }}"]`

website/docs/r/rds_instance_v3.html.markdown
```diff
@@ -0,0 +1,259 @@
...
+
+* `name` -
+  (Required)
+  Specifies the DB instance name. The DB instance name of the same type
+  must be unique for the same tenant. The value must be 4 to 64
+  characters in length and start with a letter. It is case-sensitive
+  and can contain only letters, digits, hyphens (-), and underscores
+  (_).  Changing this parameter will create a new resource.
+
+* `network_id` -
```
tstrohmeier: It would be good to mention here, that `network_id` is maped to the `subnet_id` parameter in OTC.

website/docs/r/rds_instance_v3.html.markdown
```diff
@@ -0,0 +1,259 @@
...
+
+* `availability_zone` -
+  (Required)
+  Specifies the AZ name. Changing this parameter will create a new resource.
+
+* `db` -
+  (Required)
+  Specifies the database information. Structure is documented below. Changing this parameter will create a new resource.
+
+* `flavor` -
```
tstrohmeier, line 131: Is  the argument `flavor` replacing the `flavorref` argument from v1? 


Is the following example working?
```hcl
data "opentelekomcloud_rds_flavors_v3" "flavor" {
    db_type = "PostgreSQL"
    db_version = "9.5"
    instance_mode = "ha"
}

resource "opentelekomcloud_rds_instance_v3" "instance" {
  name = "terraform_test_rds_instance"
  ...
  flavor = "${data.opentelekomcloud_rds_flavors_v3.flavor.name}"
  ...
}
```

opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
```diff
@@ -0,0 +1,1004 @@
...
+func resourceRdsInstanceV3Create(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+	client, err := config.sdkClient(GetRegion(d, config), "rdsv1", serviceProjectLevel)
+	if err != nil {
+		return fmt.Errorf("Error creating sdk client, err=%s", err)
+	}
+	client.Endpoint = strings.Replace(client.Endpoint, "/rds/v1/", "/v3/", 1)
+
+	opts := resourceRdsInstanceV3UserInputParams(d)
+	opts["region"] = GetRegion(d, config)
```
tstrohmeier, line 256: Can we still specify the region on the resource?

website/docs/r/rds_instance_v3.html.markdown
```diff
@@ -0,0 +1,259 @@
...
+    size = 100
+  }
+  flavor = "rds.pg.s1.medium.ha"
+  ha_replication_mode = "async"
+  backup_strategy = {
+    start_time = "08:00-09:00"
+    keep_days = 1
+  }
+}
+```
```
tstrohmeier, line 78: Can you please add a example where the disk encryption is used?

Am I right that I can use `opentelekomcloud_kms_key_v1` therefor?

```hcl
resource "opentelekomcloud_kms_key_v1" "key_1" {
  key_alias       = "key_1"
  realm           = "eu-de"
}

resource "opentelekomcloud_rds_instance_v3" "instance" {
 ...
  volume {
    disk_encryption_id = "${opentelekomcloud_kms_key_v1.key_1.id}"
    type = "COMMON"
    size = 100
  }
 ...
}
```

tstrohmeier (changes requested):

tstrohmeier:
related to #251

tstrohmeier:
related to #251

opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
zengchen1024: The api uses the format of "{{ availability_zone_1 }},{{ availability_zone_2 }}", not an array. If creating a single instance ,it should input one availability zone, but for primary/standy instance, it should input two available zones. see the example at https://docs.otc.t-systems.com/en-us/api/rds/en-us_topic_0160329112.html#EN-US_TOPIC_0160329112__table0863181193416

website/docs/r/rds_instance_v3.html.markdown
zengchen1024: for v1 api, the flavorref is the id of flavor. but for v3, flavorref  is the name of it.  you can not use the data source of opentelekomcloud_rds_flavors_v3 like that. because data.opentelekomcloud_rds_flavors_v3.flavor is an array, it just show all the flavors filtered by specified flavor information.

opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
zengchen1024: no need to specify it.

website/docs/r/rds_instance_v3.html.markdown
zengchen1024: you are correct.

website/docs/r/rds_instance_v3.html.markdown
zengchen1024: I will rename 'network_id' to 'subnent_id'.

niuzhenguo:
related to #257

website/docs/r/rds_instance_v3.html.markdown
tstrohmeier: Perfect and thank you for adding the section in the documentation!

website/docs/r/rds_instance_v3.html.markdown
tstrohmeier: So the recommended way is to specify the flavor with the spec code directly in the rds_v3 resource like in the example below?

```
resource "opentelekomcloud_rds_instance_v3" "instance" {
  name = "terraform_test_rds_instance"
  ...
  flavor = "rds.pg.c2.medium.ha"
  ...
}
```

opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
tstrohmeier: You are right the API specifies a string. However I think it would be a better design to specify the availability zones as a array in this TF resource. If we specify the AZs as array you can flatten this array into the string format of the API. A other benefit would be that you can check if the array only contains one AZ if no HA mode is activated and if the array contains two AZs if HA mode is active.

opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
```diff
@@ -94,13 +94,13 @@ func resourceRdsInstanceV3() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"network_id": {
```
tstrohmeier: Perfect! Thank you! This is now easier to understand.

tstrohmeier (changes requested):

website/docs/r/rds_instance_v3.html.markdown
tstrohmeier: Perfect and thank you for adding the section in the documentation!

website/docs/r/rds_instance_v3.html.markdown
zengchen1024: yes

opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
zengchen1024: agree, I will update.

website/docs/r/rds_instance_v3.html.markdown
tstrohmeier: Perfect. This makes the configuration simpler. 👍

tstrohmeier:
@zengchen1024 looks good!

zengchen1024:
@tstrohmeier thanks!!

PR #359, opened 2019-08-22
add info about bandwidth objects

dev00:
When creating FloatingIP/EIP with this module, the associated bandwidth object will be set to 1000Mbit/s. In case somebody doesn't want that, let's point to the EIP module.

Maybe deprecate the usage of `opentelekomcloud_compute_floatingip_v2` in favor of the eip module?


niuzhenguo:
@dev00 I agree that we should deprecate the usage of floatingip.

PR #382, opened 2019-10-04
Fix RDS subnet_id/network_id Confusion

r0b2g1t:
During my test I noticed, that when creating an RDS resource not the `subnet_id` is needed, but the `network_id` is required.

The fix replaces the variable `subnet_id` with `network_id`, both in the code and in the documentation.

### DIFF
--- a/opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
+++ b/opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
@@ -757,7 +757,7 @@ func buildRdsInstanceV3CreateParameters(opts map[string]interface{}, arrayIndex 
 	if e, err := isEmptyValue(reflect.ValueOf(v)); err != nil {
 		return nil, err
 	} else if !e {
-		params["network_id"] = v
+		params["subnet_id"] = v
 	}
 
 	v, err = expandRdsInstanceV3CreateVolume(opts, arrayIndex)
--- a/opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
+++ b/opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
@@ -110,7 +110,7 @@ func resourceRdsInstanceV3() *schema.Resource {
 				ForceNew: true,
 			},
 
-			"subnet_id": {
+			"network_id": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
@@ -255,7 +255,7 @@ func resourceRdsInstanceV3UserInputParams(d *schema.ResourceData) map[string]int
 		"name":                    d.Get("name"),
 		"param_group_id":          d.Get("param_group_id"),
 		"security_group_id":       d.Get("security_group_id"),
-		"subnet_id":               d.Get("subnet_id"),
+		"network_id":              d.Get("network_id"),
 		"volume":                  d.Get("volume"),
 		"vpc_id":                  d.Get("vpc_id"),
 	}
@@ -750,14 +750,14 @@ func buildRdsInstanceV3CreateParameters(opts map[string]interface{}, arrayIndex 
 		params["security_group_id"] = v
 	}
 
-	v, err = navigateValue(opts, []string{"subnet_id"}, arrayIndex)
+	v, err = navigateValue(opts, []string{"network_id"}, arrayIndex)
 	if err != nil {
 		return nil, err
 	}
 	if e, err := isEmptyValue(reflect.ValueOf(v)); err != nil {
 		return nil, err
 	} else if !e {
-		params["subnet_id"] = v
+		params["network_id"] = v
 	}
 
 	v, err = expandRdsInstanceV3CreateVolume(opts, arrayIndex)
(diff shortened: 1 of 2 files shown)


tstrohmeier:
The naming in the RDS API is `subnet_id` as mentioned here: https://github.com/terraform-providers/terraform-provider-opentelekomcloud/pull/267#discussion_r278509705
API Doc: https://docs.otc.t-systems.com/en-us/api/rds/en-us_topic_0160329112.html

In the OTC UI the `subnet.id` is displayed as `Network ID` and the `neutron_subnet_id` as `Subnet ID`. I think it would be better to talk with TSY/Huwai and clarify which ID is the "subnet id". Until then it would be better to keep this PR open since this change would break existing implementations.

r0b2g1t:
> The naming in the RDS API is `subnet_id` as mentioned here: [#267 (comment)](https://github.com/terraform-providers/terraform-provider-opentelekomcloud/pull/267#discussion_r278509705)
> API Doc: https://docs.otc.t-systems.com/en-us/api/rds/en-us_topic_0160329112.html
> 
> In the OTC UI the `subnet.id` is displayed as `Network ID` and the `neutron_subnet_id` as `Subnet ID`. I think it would be better to talk with TSY/Huwai and clarify which ID is the "subnet id". Until then it would be better to keep this PR open since this change would break existing implementations.

Thanks for your response. I will contact T-Systems/Huawei to discuss this topic. :-)

@gtema Is this possibly a topic where you can support?

tstrohmeier:
@r0b2g1t Perfect, thanks! Can you include me in the discussion please?

r0b2g1t:
@tstrohmeier I do.

tstrohmeier:
@r0b2g1t Thanks!

gtema:
@r0b2g1t, Yes, I know it's a mess - the same in otcextensions. I will try to clarify and enforce doc update

r0b2g1t:
> @r0b2g1t, Yes, I know it's a mess - the same in otcextensions. I will try to clarify and enforce doc update

Thanks a lot, do you need an request email or ticket for this topic?

gtema:
nope

tstrohmeier:
@gtema Do you have any update on this?

gtema:
well, good and bad news:

- I can confirm, that the subnet_id is expecting an ID of neutron network. It comes from the Huawei mix of FusionSphere networks and subnets. FS subnet is a neutron net. There are multiple other services (incl. CCE), which expected neutron_network_id, but the parameter is called subnet_id. It is actually likely the case in all of the FS specific services. That is also the reason, why in the UI you see neutron network_id as a subnet_id
- You can understand, nobody will do such a big change as a parameter renaming in the production system. The only chance would be to have this fixed in the next revision (API version) - and unfortunately likely this means - never.

I will try to go for escalations, but I can't promise anything.

For the moment (or better to say with respect to this particular issue) I would suggest the same, what we do in otcextensions for the CLI type - rename parameter. So end-user is passing neutron network_id, which is then passed as a subnet_id parameter into the API.

opentelekomcloud/resource_opentelekomcloud_rds_instance_v3.go
```diff
@@ -750,14 +750,14 @@ func buildRdsInstanceV3CreateParameters(opts map[string]interface{}, arrayIndex
...
-	v, err = navigateValue(opts, []string{"subnet_id"}, arrayIndex)
+	v, err = navigateValue(opts, []string{"network_id"}, arrayIndex)
 	if err != nil {
 		return nil, err
 	}
 	if e, err := isEmptyValue(reflect.ValueOf(v)); err != nil {
 		return nil, err
 	} else if !e {
-		params["subnet_id"] = v
+		params["network_id"] = v
```
gtema: I guess here it should stay subnet_id, since the API really expects subnet_id parameter, but with neutron network id content

tstrohmeier:
@gtema Thank you for clarification and the deep insights.

I just want to point out that we use the current TF RDS v3 resource in production for several systems. The change of the parameter name will maybe enforce the recreation of the DB, which MUST be avoided.

niuzhenguo:
Please follow the instruction of provider attribute renaming https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-rename

tstrohmeier:
Thank you @niuzhenguo . This would be needed here.

website/docs/r/rds_instance_v3.html.markdown
```diff
@@ -153,9 +153,9 @@ The following arguments are supported:
   Specifies the security group which the RDS DB instance belongs to.
   Changing this parameter will create a new resource.
 
-* `subnet_id` -
+* `network_id` -
```
tstrohmeier, line 156: As pointed out by @niuzhenguo there should be a smooth migration.
https://www.terraform.io/docs/extend/best-practices/deprecations.html#renaming-a-required-attribute

* `subnet_id` 
**Deprecated**
(Optional)
Specifies the subnet id. Changing this parameter will create a new resource. Either subnet id or network id MUST be specified.
* `network_id ` 
(Optional)
Specifies the network id. Changing this parameter will create a new resource. Either subnet id or network id MUST be specified.

tstrohmeier (changes requested):

website/docs/r/rds_instance_v3.html.markdown
tstrohmeier: @r0b2g1t Why is this conversation marked as resolved?

tstrohmeier:
@r0b2g1t As pointed out by niuzhenguo there should be a smooth migration.
https://www.terraform.io/docs/extend/best-practices/deprecations.html#renaming-a-required-attribute

* `subnet_id` 
**Deprecated**
(Optional)
Specifies the subnet id. Changing this parameter will create a new resource. Either subnet id or network id MUST be specified.
* `network_id ` 
(Optional)
Specifies the network id. Changing this parameter will create a new resource. Either subnet id or network id MUST be specified.

outcatcher:
Hi @r0b2g1t, this PR is stale for 7 months already. Do you still need it?

PR #386, opened 2019-10-11
Fix admin_state_up bool required Issue for ELB Load Balancer

r0b2g1t:
The pull request fixes the documentation issue for the variable `admin_state_up` of "opentelekomcloud_elb_loadbalancer". By correcting the use of Booleans for variable `admin_state_up`, instead of the values `1` or `0` as documented.

The error was shown by Terraform when I planed my infrastructure:

Code:
```
...
resource "opentelekomcloud_elb_loadbalancer" "frontend-elb" {
  name           = "frontend-elb"
  type           = "Internal"
  description    = "ELB of the frontend instances"
  vpc_id         = "${opentelekomcloud_networking_network_v2.tier-3.id}"
  admin_state_up = 1
}
...

```
Error:
```
Error: Incorrect attribute value type

  on modules/network/main.tf line 92, in resource "opentelekomcloud_elb_loadbalancer" "frontend-elb":
  92:   admin_state_up = 1

Inappropriate value for attribute "admin_state_up": bool required.
```

### DIFF
--- a/website/docs/r/elb_loadbalancer.html.markdown
+++ b/website/docs/r/elb_loadbalancer.html.markdown
@@ -18,7 +18,7 @@ resource "opentelekomcloud_elb_loadbalancer" "elb" {
   type = "External"
   description = "test elb"
   vpc_id = "e346dc4a-d9a6-46f4-90df-10153626076e"
-  admin_state_up = 1
+  admin_state_up = "true"
   bandwidth = 5
 }
 ```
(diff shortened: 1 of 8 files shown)


website/docs/r/elb_loadbalancer.html.markdown
```diff
@@ -45,9 +45,9 @@ The following arguments are supported:
     Internal or External. Changing this creates a new elb loadbalancer.
 
 * `admin_state_up` - (Required) Specifies the status of the load balancer.
-    Value range: 0 or false: indicates that the load balancer is stopped. Only
-    tenants are allowed to enter these two values. 1 or true: indicates that
-    the load balancer is running properly. 2 or false: indicates that the load
+    Value range: false: indicates that the load balancer is stopped. Only
+    tenants are allowed to enter these two values. true: indicates that
+    the load balancer is running properly. false: indicates that the load
     balancer is frozen. Only tenants are allowed to enter these two values.
```
niuzhenguo: Can we refine this with "Value range: false: indicates that the load balancer is stopped or frozen; true: indicates that the load balancer is running properly. Only tenants are allowed to enter these two values."

PR #433, opened 2019-12-17
Updated list of values for volume_type

CSchmidtDD:
The original documentation does not refer to volumes, so the documentation needs to be updated accordingly

Sadly, the Volume Creation in Line 88 on [resource_opentelekomcloud_blockstorage_volume_v2.g](https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/opentelekomcloud/resource_opentelekomcloud_blockstorage_volume_v2.go) lacks the validation changes made in [evs_volume_v3](https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/website/docs/r/evs_volume_v3.html.markdown), that is referred, might pull errors that confuse.

Also it is not clear what every Option does, there is only Trial&Error.
Maybe order from Common I/O to High Performance I/O and Low Latency I/O

### DIFF
--- a/website/docs/r/blockstorage_volume_v2.html.markdown
+++ b/website/docs/r/blockstorage_volume_v2.html.markdown
@@ -68,7 +68,7 @@ The following arguments are supported:
 * `source_vol_id` - (Optional) The volume ID from which to create the volume.
     Changing this creates a new volume.
 
-* `volume_type` - (Optional) Currently, the value can be `SSD` (ultra-high I/O disk type), `SAS` (high I/O disk type), `SATA` (common I/O disk type), `co-p1` (Excluise HPC/ SAP HANA: high I/O performance optimized), or `uh-l1` (Excluise HPC/ SAP HANA: ultra-high-I/O latency optimized), Read Note for `uh-l1` and `co-p1`: [OTC-API](https://docs.otc.t-systems.com/en-us/api/ecs/en-us_topic_0065817708.html). Changing this creates a new volume.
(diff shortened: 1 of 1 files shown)


website/docs/r/blockstorage_volume_v2.html.markdown
```diff
@@ -69,7 +69,7 @@ The following arguments are supported:
     Changing this creates a new volume.
 
 * `volume_type` - (Optional) The type of volume to create.
-    Changing this creates a new volume.
+    Changing this creates a new volume. Currently, the value can be SSD, SAS, SATA, co-p1, or uh-l1. Changing this creates a new volume.
```
ShiChangkuo: "Changing this  creates a new volume." was duplicated.

ShiChangkuo:
@Chrakker currently, the verification for some parameters are not sufficient. thanks for your efforts, please see my comments inline.

website/docs/r/blockstorage_volume_v2.html.markdown
CSchmidtDD: usually, the documentation specifies under every attribute, which changes lead to a new creation. I can not see a duplication here. I would change the current working to the new volume - Update documentatio:
(Optional) Currently, the value can be SSD (ultra-I/O disk type), SAS (high I/O disk type), or SATA (common I/O disk type) OTC-API

CSchmidtDD: The duplication should be a non-issue now, updated the documentation to a more extensive and readable way. 
/Resolved

CSchmidtDD:
> 
> 
> The original documentation does not refer to volumes, so the documentation needs to be updated accordingly
> 
> Sadly, the Volume Creation in Line 88 on [resource_opentelekomcloud_blockstorage_volume_v2.g](https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/opentelekomcloud/resource_opentelekomcloud_blockstorage_volume_v2.go) lacks the validation changes made in [evs_volume_v3](https://github.com/terraform-providers/terraform-provider-opentelekomcloud/blob/master/website/docs/r/evs_volume_v3.html.markdown), that is referred, might pull errors that confuse.
> 
> Also it is not clear what every Option does, there is only Trial&Error.
> Maybe order from Common I/O to High Performance I/O and Low Latency I/O

In the updated documentation as of now, the options are clear what does what. It is also referenced to the API which option can be used (note added) and what can be referenced to billing.

PR #434, opened 2019-12-17
Add `clouds.yaml` support

outcatcher:
Support of  `clouds.yaml` works both in `huaweicloud` and `openstack` providers

This PR introduces support of `clouds.yaml` for `opentelekomcloud` provider using approach from providers mentioned above with respect of environment variables handling described in readme

Also there is documentation changes describing `clouds.yaml` support


website/docs/index.html.markdown
```diff
@@ -32,6 +32,14 @@ resource "opentelekomcloud_compute_instance_v2" "test-server" {
 }
 ```
 
+Or using `clouds.yaml` file:
+```hcl
+provider "opentelekomcloud" {
+  cloud = var.cloud_name
+}
+# ...
+```
```
niuzhenguo: Better to move this to Authentication section below "This provider offers 4 means for authentication."

outcatcher: Moved clouds.yaml usage to auth section. Though I'm not sure, if some more information should be added

outcatcher:
I had some conflicts during rebase onto `upstream/master`. Sorry for stupid question, but maybe there is something I can do to to avoid reordering of dependencies during `go mod vendor`, as it adds a lot of useless changes in `vendor/modules.txt`?

niuzhenguo:
@outcatcher that's OK.

PR #460, opened 2020-03-04
Fix vpc_subnet_v1 retrieval by id

dlandtwing:
I've noticed that when trying to obtain a vpc_subnet_v1 data source by id as follows, the following error occurs:
```
data "opentelekomcloud_vpc_subnet_v1" "test-vpc-subnet" {
  id = "7c1aaf60-bf66-4a53-b823-fd07192bf18d"
}
```
error message:
`Error: multiple subnets matched; use additional constraints to reduce matches to a single subnet`

The "id" attribute seems to be ignored for filtering subnets. With this change, the above subnet ist retrieved correctly


niuzhenguo:
Thanks for the contribution!

PR #462, opened 2020-03-05
Fix removing FW rule assigned to FW policy

### LINKED ISSUE #436: Can't delete firewall rule(s)
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.12.12
+ provider.local v1.4.0
+ provider.opentelekomcloud v1.13.1
+ provider.template v2.1.2

### Affected Resource(s)
- opentelekomcloud_fw_rule_v2

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
  # opentelekomcloud_fw_rule_v2.default-ssh will be destroyed
  - resource "opentelekomcloud_fw_rule_v2" "default-ssh" {
      - action                 = "deny" -> null
      - destination_ip_address = "0.0.0.0/0" -> null
      - enabled                = true -> null
      - id                     = "3dcec8de-ff7a-4535-bbb3-d7f208752bdd" -> null
      - ip_version             = 4 -> null
      - protocol               = "tcp" -> null
      - region                 = "eu-de" -> null
      - source_ip_address      = "0.0.0.0/0" -> null
    }
```

### Debug Output
https://gist.github.com/nicinuse/79696b4223289011fd75144bf1296d25

### Expected Behavior
Firewall rule should be deleted, maybe disabled before.

### Actual Behavior
Firewall is not deleted, because it is enabled.

### Steps to Reproduce
`terraform apply`

outcatcher:
Fix #436

Tested with firewall configuration provided in #436: failing in base version, passing in new

Added acceptance test to cover the issue

### DIFF
--- a/opentelekomcloud/resource_opentelekomcloud_fw_rule_v2.go
+++ b/opentelekomcloud/resource_opentelekomcloud_fw_rule_v2.go
@@ -8,6 +8,7 @@ import (
 	"github.com/huaweicloud/golangsdk"
 	"github.com/huaweicloud/golangsdk/openstack/networking/v2/extensions/fwaas_v2/policies"
 	"github.com/huaweicloud/golangsdk/openstack/networking/v2/extensions/fwaas_v2/rules"
+	"github.com/huaweicloud/golangsdk/pagination"
 )
 
 func resourceFWRuleV2() *schema.Resource {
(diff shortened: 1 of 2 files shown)


opentelekomcloud/resource_opentelekomcloud_fw_rule_v2.go
```diff
@@ -246,6 +251,30 @@ func resourceFWRuleV2Delete(d *schema.ResourceData, meta interface{}) error {
 	return rules.Delete(networkingClient, d.Id()).Err
 }
 
+func assignedPolicyID(nwClient *golangsdk.ServiceClient, ruleID string) (string, error) {
```
niuzhenguo, line 254: Why we need to check this? Do you mean the policy update method not take effect for the last rule?

outcatcher: The problem, rule can be removed before policy is updated (and it always does), so we should be sure to remove rule from policy.
I haven't change the logic of `resourceFWRuleV2Delete`, if was intended to work the same way before. Previously there was a check for `rule.PolicyID` to be not empty, but it is not set anywhere and not retrieved using `rules.Get`, so search for policies is required. This solution seems to be ugly from my point of view, but I can't find any other way to do that.

niuzhenguo: So, it should be a BUG on server side right? rule.PolicyID should be set there but in fact not.

outcatcher: No, it is not an issue ccording VPC spec, there is no `policy_id` in response: https://docs.otc.t-systems.com/api/vpc/vpc_firewall_0002.html

PR #470, opened 2020-03-23
Update networking_floatingip_associate_v2.html.markdown

shaderecker:
There is also the ability to assign the Floating IP to an instance


shaderecker:
Nvm, confused it with https://www.terraform.io/docs/providers/opentelekomcloud/r/compute_floatingip_associate_v2.html

PR #496, opened 2020-05-11
fix waf_domain_v1 using old waf api

schrej:
The server props were changed from front_ and back_protocol
to client_ and server_protocol without changing api version.

This depends on huaweicloud/golangsdk#253 and will probably require a change to go.mod after that's merged.

### DIFF
--- a/opentelekomcloud/resource_opentelekomcloud_waf_domain_v1.go
+++ b/opentelekomcloud/resource_opentelekomcloud_waf_domain_v1.go
@@ -237,8 +237,8 @@ func resourceWafDomainV1Read(d *schema.ResourceData, meta interface{}) error {
 	servers := make([]map[string]interface{}, len(n.Server))
 	for i, server := range n.Server {
 		servers[i] = make(map[string]interface{})
-		servers[i]["client_protocol"] = server.FrontProtocol
-		servers[i]["server_protocol"] = server.BackProtocol
+		servers[i]["client_protocol"] = server.ClientProtocol
+		servers[i]["server_protocol"] = server.ServerProtocol
 		servers[i]["address"] = server.Address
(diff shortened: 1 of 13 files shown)


niuzhenguo:
@schrej Thanks for the patch, golangsdk changes already been merged. Please do an update here.

opentelekomcloud/resource_opentelekomcloud_waf_domain_v1.go
```diff
@@ -45,11 +45,11 @@ func resourceWafDomainV1() *schema.Resource {
 				ForceNew: false,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"front_protocol": {
```
niuzhenguo, line 48: How about keeping this name and just modifying the SDK part, so the new version of provider will not break users' original setup.

schrej: The best way would probably be to deprecate the old ones.
https://www.terraform.io/docs/extend/best-practices/deprecations.html#renaming-a-required-attribute

niuzhenguo: yes, the best way is to deprecate the old ones.

niuzhenguo: I will do a quick fix for this as this blocks others update the golangsdk.

schrej: Ok. We should still rename the parameters to be consistent with the API though. I can work on that, but it'll probably take a while until I find time to do it. I've created #509 to keep track of that.

PR #510, opened 2020-05-22
make tags as a common method in provider

ShiChangkuo:


iaean:
👍 Possibility to tag all resources is mandatory, especially if someone needs to implement cost control. Thanks a lot for adding this improvement.

Issue #530, opened 2020-06-29
[Feature Request] CTS Tracker per project


martingbl:
Hi there,
please provide the possibility to set the project for a CTS tracker.

### Terraform Version
Terraform v0.12.26

### Affected Resource(s)
- resource_opentelekomcloud_cts_tracker_v1.go 
- and everything affected https://github.com/terraform-providers/terraform-provider-opentelekomcloud/pull/135/files

### Expected Behavior
I expect to be able to provide a project name to the tracker (one tracker named system can be set per project)

### Actual Behavior
Only adds a tracker to the base project. E.g. `eu-de`

### References
- GH-135

outcatcher:
@TheTaka96 can you please clarify your use case? Do you want to create trackers for multiple child projects authorizing with credentials for the base project?

martingbl:
@outcatcher okay so the project named after the region will be described as region in the following and all "subprojects" will be called projects:

E.g. we have a region named `eu-de` in which we want several projects. For example `production` and `preproduction`.
Now those projects need a tracker to log changes in the environment (e.g. ECS creation) using an OBS Bucket.

The tracker on region level does not display the according information for the individual problems but it is the default value right now in Terraform.

I hope this was clear enough. Just adding and using a project variable on resource creation for the request url should be enough to fulfill this usecase.

PR #554, opened 2020-07-27
Add project_name to SMN topic

lego963:
Part of: #530

### DIFF
--- a/opentelekomcloud/provider.go
+++ b/opentelekomcloud/provider.go
@@ -472,8 +472,8 @@ func configureProvider(d *schema.ResourceData, terraformVersion string) (interfa
 }
 
 func reconfigProjectName(src *Config, projectName string) (*Config, error) {
-	config := Config{}
-	err := copier.Copy(&config, &src)
+	config := &Config{}
+	err := copier.Copy(config, src)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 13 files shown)


opentelekomcloud/provider.go
```diff
@@ -469,3 +470,16 @@ func configureProvider(d *schema.ResourceData, terraformVersion string) (interfa
 
 	return &config, nil
 }
+
+func reconfigProjectName(src Config, projectName string) (*Config, error) {
```
outcatcher: `src` type should be `&Config`, `config` type should be `&Config` too, as they are used as pointers

opentelekomcloud/util.go
```diff
@@ -49,6 +49,20 @@ func GetRegion(d *schema.ResourceData, config *Config) string {
...
 }
 
+// GetProjectName returns the project name that was specified in the resource.
+func GetProjectName(d *schema.ResourceData, config *Config) string {
+	projectNameField := d.Get("project_name")
+	value := projectNameField.(string)
+	if value != "" {
+		return value
+	}
+	n := config.TenantName
```
outcatcher: Please don't name var `n`

outcatcher (changes requested):

outcatcher (approved):

PR #558, opened 2020-07-31
Add opentelekomcloud_compute_availability_zones_v2 data source

### LINKED ISSUE #528: [Compute] opentelekomcloud_compute_availability_zones_v2 data source is missing
Provider has no `opentelekomcloud_compute_availability_zones_v2` data source, that forces user to use `openstack_compute_availability_zones_v2` instead.

To decrease number of used providers, this data source should be implemented in OTC provider too.

**Provider version**: git master branch

outcatcher:
Fixes #528


opentelekomcloud/data_source_opentelekomcloud_compute_availability_zones_v2.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	stateBool := d.Get("state").(string) == "available"
+	zones := make([]string, 0, len(zoneInfo))
+	for _, z := range zoneInfo {
+		if z.ZoneState.Available == stateBool {
+			zones = append(zones, z.ZoneName)
+		}
+	}
+
```
lego963, line 65: I guess that we need to check if we return at least 1 zone

lego963 (approved):

Issue #562, opened 2020-08-07
rds tags not working after switchover for rds ha instance 


otc-fabo:
when using the rds in ha mode together with tags, there is a problem in TF after performing a switchover between primary/standby instance. 
Background information from OTC:
Tags are only for the primary node.

### Terraform Version
Terraform v0.12.28
+ provider.opentelekomcloud v1.18.1

### Affected Resource(s)
opentelekomcloud_rds_instance_v3 

### Terraform Configuration Files
```hcl
resource "opentelekomcloud_rds_instance_v3" "instance" {
  availability_zone = ["eu-de-03","eu-de-02"]
  db {
    password = var.pw
    type = "PostgreSQL"
    version = "11"
    port = "8635"
  }
  name = "terraform_test_rds_instance"
  security_group_id = var.sec_group
  subnet_id = var.subnet
  vpc_id = var.vpc
  volume {
    type = "COMMON"
    size = 100
  }
  flavor = "rds.pg.c2.large.ha"
  backup_strategy {
    start_time = "08:00-09:00"
    keep_days = 1
  }
  ha_replication_mode = "async"
  tag = {
    foo = "bar"
    key = "value"
  }
}

```

### Debug Output
Error: Error fetching OpenTelekomCloud rds instance tags: Expected HTTP response code [200] when accessing [GET https://rds.eu-de.otc.t-systems.com/v1/c87a0390a7f242a8b32b104b4e899793/rds/887b4261fb894b729d0b13ab0b6cad8cno03/tags], but got 415 instead
{"errCode":"RDS.0004","externalMessage":"Invalid operation."}

### Expected Behavior
terraform to work as expected

### Actual Behavior
terraform to recognize the switchover of the primary/standby instance 

### Steps to Reproduce
1. create rds ha instance with tags
2. login to OTC -> RDS -> Instance -> Switch primary/standby
3. add another tag to the tf-code
4. run terraform apply again

### References
same error as in https://github.com/terraform-providers/terraform-provider-opentelekomcloud/issues/449

outcatcher:
The error should no more appear in 1.19.1

PR #564, opened 2020-08-13
Fix HTTP 415 when retrieving tags after nodes role switch

outcatcher:
Make provider use `master` node instead of node with ID `_node0` for getting tags


outcatcher:
Possible fix of #562 

HTTP 415 is not reproduced any more, though correct work of provider with switched RDS nodes requires setting
```hcl
  lifecycle {
    ignore_changes = [
      "db",
      "availability_zone",
    ]
  }
```
So terraform won't treat switching nodes as a change in state ― availability zones got reordered, and change in AZ forces re-creation of RDS instance by default.

niuzhenguo (approved):

PR #569, opened 2020-08-18
Remove `vendor` directory

outcatcher:
Is there any reason to keep `vendor` directory?
@niuzhenguo, @ShiChangkuo, is it okay to remove it?


niuzhenguo:
The vendor directory stayed after we moved to Go Modules is because it doesn't require much effort to maintain it and can help old version of go env work, no special reason to my understand, but seems all other providers still keep it now.

outcatcher:
The problems from my PoV ― there is a lot of changes when adding new dependencies. Also, Linux and Windows add system-specific libraries on every `go mod vendor`. For building the provider there is requirement go 1.11+, so modules should work anyway. I've asked in TF [gitter](https://gitter.im/hashicorp-terraform/Lobby) about that :) Maybe guys from there will give some hint about `vendor`

PR #575, opened 2020-08-21
add goreleaser-build job

gtema:
Depends-On: https://github.com/opentelekomcloud-infra/otc-zuul-jobs/pull/16

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -15,11 +15,13 @@
         - golang-make-test
         - golang-make-vet
         - golang-make-website-test
+        - goreleaser-build
     gate:
       jobs:
         - golang-make-test
         - golang-make-vet
         - golang-make-website-test
+        - goreleaser-build
     tag:
       jobs:
         - release-goreleaser


.zuul.yaml
```diff
@@ -15,11 +15,13 @@
         - golang-make-test
         - golang-make-vet
         - golang-make-website-test
+        - goreleaser-build
     gate:
       jobs:
         - golang-make-test
         - golang-make-vet
         - golang-make-website-test
+        - goreleaser-build
```
outcatcher, line 24: Why `goreleaser-build` but `release-goreleaser` in tag jobs?

gtema: that can be in a followup, but
all our real release jobs are "release-xxx" (release-python, release-ansible-collection). We can say "release with the help of goreleaser". Other jobs (or actually all) are following structure of the command: in the golang execute make test, in the goreleaser invoke build, ...

outcatcher (approved):

PR #583, opened 2020-08-24
do not use armor for signing

gtema:
This solves provider publishing (at least after manual resigning without --armor and reuploading signature solved it immediately)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -40,7 +40,6 @@ signs:
     args:
       - "--output"
       - "${signature}"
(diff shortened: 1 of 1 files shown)


outcatcher:
Maybe we also should add selecting the exact private key for signing? Or used VM has only one pgp secret key?
```
      - "--local-user"
      - "{{ .Env.GPG_FINGERPRINT }}" # set this environment variable for your signing key
```

outcatcher (approved):

Issue #588, opened 2020-08-26
[GNUMakefile] website-test


lego963:
Since https://github.com/hashicorp/terraform-website/commit/03518a8e7cf7efc724c46acfc617ebbb3a2c4593 we have to fix our makefile

gtema:
I guess we need then to reprioritize https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/issues/577.

The website job is not required at all

Issue #577, opened 2020-08-24
Migrate provider docs from website/**md.html into docs/**.md


gtema:
We need to start slowly migrating docs from the current format into the new as described in https://www.terraform.io/docs/registry/providers/docs.html

PR #589, opened 2020-08-27
Remove not required jobs

lego963:
Refers to: #577 and #588

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ install:
 script:
   - make test
   - make vet
-#- make website-test
-#NEED TO BE FIXED: in #577, #588
 
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,33 @@
 dist: trusty
 sudo: required
 services:
-- docker
+  - docker
 language: go
 go:
-- "1.11.x"
+  - "1.11.x"
 
 install:
-# This script is used by the Travis build to install a cookie for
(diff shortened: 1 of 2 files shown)


gtema:
I would say it's not temporary, in #577 we should drop them as such and migrate docs

.travis.yml
```diff
@@ -1,32 +1,33 @@
...
+  # See: https://github.com/golang/go/issues/12933
+  - bash scripts/gogetcookie.sh
 
 script:
-- make test
-- make vet
-- make website-test
+  - make test
+  - make vet
+#- make website-test
```
outcatcher: Please remove it completely. Don't comment out unused things

.zuul.yaml
```diff
@@ -14,13 +14,15 @@
       jobs:
         - golang-make-test
         - golang-make-vet
-        - golang-make-website-test
+        #- golang-make-website-test
```
outcatcher: See prev comment

.zuul.yaml
```diff
@@ -14,13 +14,15 @@
...
+        #NEED TO BE FIXED: in #577, #588
         - goreleaser-build
     gate:
       jobs:
         - golang-make-test
         - golang-make-vet
-        - golang-make-website-test
+        #- golang-make-website-test
```
outcatcher: See prev comment

outcatcher (changes requested):

outcatcher (approved):

PR #593, opened 2020-08-27
Docs migrations

### LINKED ISSUE #577: Migrate provider docs from website/**md.html into docs/**.md

lego963:
Resolve: #577


gtema (changes requested):
Please go through the readme and fix links to old repo location (unfortunately GH does not allow commenting lines not touched by PR)

gtema:
you should not remove the "--- frontmatter ---" headers from files. At least page_title should stay (actually perhaps start using subcategory). Read carefully https://www.terraform.io/docs/registry/providers/docs.html

lego963:
Yes, I will return it, now I format the documents and fix them with strange or insufficient info

docs/data-sources/antiddos_v1.md
```diff
@@ -1,63 +1,59 @@
...
-
-* `traffic_cleaning_status` - The traffic cleaning status.
-
-* `trigger_bps` - The traffic at the triggering point.
-
-* `trigger_pps` - The packet rate at the triggering point.
-
-* `trigger_http_pps` - The HTTP request rate at the triggering point.
+---
+subcategory: "AntiDDos"
```
outcatcher: It should be `DDoS`, not `DDos`

gtema (dismissed):
Whew.
I don't know whether we will get problems with removal of `title`, but we can re-add it then.

outcatcher (changes requested):

lego963:
>`page_title - The title of this document, which will display in the docs navigation. This is only required for documents in the guides/ folder.`

according to https://www.terraform.io/docs/registry/providers/docs.html

gtema:
> > `page_title - The title of this document, which will display in the docs navigation. This is only required for documents in the guides/ folder.`
> 
> according to https://www.terraform.io/docs/registry/providers/docs.html

it's not about "requires", it's about how the page name would be shown by the render tool, but even before it was not very much useful. On the other hand all other providers do prefix pages with the provider name (like it it currently)

gtema:
hmm, the build is also failing in a weird way

outcatcher (dismissed):

gtema (approved):
No clue how to review 271 files other that simply by diagonal

PR #621, opened 2020-09-21
add k8sTags to cce node resource

### LINKED ISSUE #552: CCE node support K8s labels
please upgrade the TF Provider for opentelekomcloud_cce_node_v3 
https://www.terraform.io/docs/providers/opentelekomcloud/r/cce_node_v3.html
provider.opentelekomcloud v1.18.1

to support Kubernetes labels during the creation of a node. 
Currently, there are two parameter for node creation: 

labels - (Optional) Node tag, key/value pair format. Changing this parameter will create a new resource.

tags - (Optional) The field is alternative to labels, key/value pair format.

labels had no function at all (tested with node creation for clusters running: v1.13.10-r0 | v1.11.7-r2 | v1.15.6-r1).
tags is working fine and set the OTC tags. 
It is not possible to use both parameter at the same time. 

please either fix labels and use it to set the kubernetes tags or create a new parameter for it. 
Rest API is supporting it:
parameter: k8sTags
https://docs.otc.t-systems.com/en-us/api2/cce/cce_02_0242.html

### Affected Resource(s)
opentelekomcloud_cce_node_v3

cgroschupp:
Allow to add k8s tags.

https://github.com/huaweicloud/golangsdk/blob/master/openstack/cce/v3/nodes/results.go#L75

Fixes #552

### DIFF
--- a/opentelekomcloud/resource_opentelekomcloud_cce_node_v3.go
+++ b/opentelekomcloud/resource_opentelekomcloud_cce_node_v3.go
@@ -5,9 +5,10 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"fmt"
-	"github.com/hashicorp/go-multierror"
 	"log"
 	"time"
+
+	"github.com/hashicorp/go-multierror"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -254,6 +255,11 @@ func resourceCCENodeV3() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"k8s_tags": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


outcatcher (changes requested):
Thanks for your contribution to the OTC Terraform provider! 

The only remark ― can you please also implement validation for `k8sTags` as described in the documentation:
>Key: Enter 1 to 63 characters starting with a letter or digit. Only letters, digits, hyphens (-), underscores (_), and periods (.) are allowed. A DNS subdomain can be prefixed to a key and contain a maximum of 253 characters. Example DNS subdomain: example.com/my-key
>Value: The value can be left blank or a string of 1 to 63 characters starting with a letter or digit. Only letters, digits, hyphens (-), underscores (_), and periods (.) are allowed in the character string.

opentelekomcloud/resource_opentelekomcloud_cce_node_v3.go
```diff
@@ -16,6 +17,47 @@ import (
...
+				"key %q cannot be longer than 63 characters: %q", k, key))
+		}
+
+		if len(valueString) > 63 {
+			errors = append(errors, fmt.Errorf(
+				"value %q cannot be longer than 63 characters: %q", k, value))
+		}
+
+		pattern := `^[\.\-_A-Za-z0-9]+$`
+		if !regexp.MustCompile(pattern).MatchString(key) {
```
outcatcher: There are duplicating operations for each iteration.
Please make `pattern` a const and move `regexp.MustCompile(pattern)` out of the `for` cycle

outcatcher (changes requested):
Sorry for the slow response ― there are issues with CCE tests

outcatcher:
@cgroschupp would you like to continue with the PR? You need to resolve conflict with `master` branch now.

outcatcher:
@cgroschupp There are conflicts with the base branch. Can you please merge `master` into your branch first (or rebase your branch onto up-to-date master)?

cgroschupp:
@outcatcher sorry for the delay

lego963 (approved):

outcatcher (approved):
Tests passed

PR #1058, opened 2021-05-17
[VPC] Query EIPs by tags

### LINKED ISSUE #1022: Provide data sources for opentelekomcloud_vpc_eip_v1
# Description of the enhancement
It would be nice if terraform [data sources](https://www.terraform.io/docs/language/data-sources/index.html) would be supported for `opentelekomcloud_vpc_eip_v1`. This way more advanced configurations which consist of multiple moving parts could be achieved more easily. 

Example: 

For a consistent deployment a static IP should (almost) never be destroyed. 
Currently there are to options: 
- Either created manually before terraform deployment and manually provide the ip address 
- or create it via terraform and the manually remove it from the state

With `data sources` one could create the IP (or other one time resources) with one terraform deployment and the create another deployment which references the previously created resources by selecting its name.

```hcl
data "opentelekomcloud_vpc_eip_v1" "lb" {
  name = "${var.ROOT_NAME}-cce-cluster-public-ip"
}
```
Thus one can control the one time resources with terraform instead of having any manually steps in-between.

lego963:
Add possibility to query EIPs by tags in `data_source/opentelekomcloud_vpc_eip_v1`

Resolves: #1022

## PR Checklist

* [x] Refers to: #1022
* [x] Tests added/passed.
* [x] Documentation updated.
* [x] Schema updated.

## Acceptance Steps Performed

```
=== RUN   TestAccVpcEipV1DataSource_basic
--- PASS: TestAccVpcEipV1DataSource_basic (128.45s)
PASS

```

### DIFF
--- a/opentelekomcloud/services/vpc/data_source_opentelekomcloud_vpc_eip_v1.go
+++ b/opentelekomcloud/services/vpc/data_source_opentelekomcloud_vpc_eip_v1.go
@@ -3,10 +3,13 @@ package vpc
 import (
 	"fmt"
 	"log"
+	"reflect"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/opentelekomcloud/gophertelekomcloud/openstack/common/tags"
 	"github.com/opentelekomcloud/gophertelekomcloud/openstack/networking/v1/eips"
+	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/common"
 
 	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/common/cfg"
 )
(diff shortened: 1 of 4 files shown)


opentelekomcloud/services/vpc/data_source_opentelekomcloud_vpc_eip_v1.go
```diff
@@ -124,9 +158,29 @@ func dataSourceVPCEipV1Read(d *schema.ResourceData, meta interface{}) error {
...
+	)
+
 	if mErr.ErrorOrNil() != nil {
 		return mErr
 	}
 
 	return nil
 }
+
+func contains(s []tags.ResourceTag, tag tags.ResourceTag) bool {
```
outcatcher: Please check if `tags.ResourceTag` can be replaced with `interface{}` here

outcatcher (changes requested):

outcatcher (approved):

PR #1164, opened 2021-06-25
Initialize releasenotes building

gtema:
- setup releasenotes structure to be build by sphinx&reno (same as other python otc tools)
- migrate content of the changelog into older releasenotes

=======
- RN publishing is going to be implemented separately in the job-template (bit later)
- PR's that have interface effect "must" contain RN note (can be generated according to https://docs.openstack.org/reno/latest/user/usage.html)
- PR's without external visibility "should" have RN note.
- it is in the reviewers responsibility to check/enforce RN presence. Generally RN can be added later, but should be done before release, cause otherwise will be related to wrong release.

### DIFF
--- a/releasenotes/source/current.rst
+++ b/releasenotes/source/current.rst
@@ -1,0 +1,61 @@
+=====================
+Current Release Notes
+=====================
+
+.. release-notes::
+
+1.24.2
+------
+
+New Features
+============
+
+* `resource/opentelekomcloud_waf_certificate_v1`: Add possibility to configure TLS ([#1132](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/issues/1132))
+* `resource/opentelekomcloud_ecs_instance_v1`: Add possibility to encrypt `data_volumes` ([#1135](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/issues/1135))
+
+Bug Fixes
+=========
+
+* `provider`: Store temporary AK/SK in OBS client only ([#1133](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/issues/1133))
+* `resource/opentelekomcloud_rds_instance_v3`: Fix not setting `availability_zone` during import/refresh ([#1137](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/issues/1137))
+* `resource/opentelekomcloud_obs_bucket`: Fix not creating bucket for `eu-nl` region ([#1139](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/issues/1139))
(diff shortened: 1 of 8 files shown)


gtema:
Job artifact docs-html.tar.gz that represent result. Please look at it and see whether you want some restructure or it is fine this way.

lego963 (approved):

releasenotes/source/current.rst
```diff
@@ -0,0 +1,61 @@
...
+* **New Resource:** `opentelekomcloud_swr_repository_v2` ([#1123](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/pull/1123))
+* **New Resource:** `opentelekomcloud_swr_domain_v2` ([#1127](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/pull/1127))
+* **New Resource:** `opentelekomcloud_swr_organization_permissions_v2` ([#1129](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/pull/1129))
+* `resource/opentelekomcloud_waf_certificate_v1`: Add name certificate name validation ([#1116](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/pull/1116))
+* `resource/opentelekomcloud_cce_node_v3`: Add possibility to encrypt `data_volumes` ([#1117](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/pull/1117))
+* `resource/opentelekomcloud_waf_certificate_v1`: Add possibility to import by name ([#1119](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/pull/1119))
+* `resource/opentelekomcloud_obs_bucket_v1`: Add possibility to create buckets with encryption ([#1125](https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/pull/1125))
+
```
outcatcher, line 43: Can we have new features and existing resource enhancements separated, as it was before?

gtema: https://docs.openstack.org/reno/latest/user/usage.html#editing-a-release-note
reno is having list of supported sections (which doesn't differentiate those). There is generally no real difference between new features and enhancements

outcatcher: Just a question of grouping, really. Format of `**New Resource:**` thing is bit out of line in this case.

gtema: it is possible to add new sections (https://docs.openstack.org/reno/latest/user/usage.html#configuring-reno), but do you thing this is really required? I personally do not see difference between adding new resource and adding new parameter to existing resource

gtema: actually this was only for importing old RN. I first tried to convert them under default reno style, but then gave up cause it's too much manual work

gtema: > Just a question of grouping, really. Format of `**New Resource:**` thing is bit out of line in this case.

we can drop it for old entries, or replace it with "add XXX"

outcatcher (approved):

Issue #1151, opened 2021-06-23
[CCE] Add taints to CCE node


0xd3e:
# Description of the enhancement

I've seen that it is possible to add taints to a CCE node pool, but not to a CCE node.
The API seems to support it.

I'd be happy to write a PR if you're interested.

outcatcher:
@0xIDANT we are always happy then people contribute to the provider 🙂 Please write a PR

0xd3e:
@outcatcher I've realized that I could possibly implement providing taints for node creation, but the API doesn't allow updating the node specs. Only updating the node name is supported by the OTC API. So updating node taints won't work with the current API.

This is a bit weird because the node pool supports updating the taints.

Should I go forward and implement it for node creation only?

outcatcher:
Let's start with creation-only :) marked with `ForceNew`

outcatcher:
Internal incident for no taints in the response:
https://jira.tsi-dev.otc-service.com/browse/OCS-1477

lego963:
Fixed in: #1191

PR #1191, opened 2021-07-02
Add option to add taints to resource_opentelekomcloud_cce_node_v3

0xd3e:

This PR adds the option to add Kubernetes taints to the resource opentelekomcloud_cce_node_v3.

## PR Checklist

* [x] Refers to: #1151 
* [x] Tests added/passed.
* [x] Documentation updated.
* [x] Schema updated.

## Acceptance Steps Performed

I didn't run the acceptance tests, because I do not have an OTC account that I could use for this, but I did add a test that could be executed.

### DIFF
--- a/docs/resources/cce_node_v3.md
+++ b/docs/resources/cce_node_v3.md
@@ -65,6 +65,11 @@ The following arguments are supported:
 
 * `annotations` - (Optional) Node annotation, key/value pair format. Changing this parameter will create a new resource.
 
+* `taints` - (Optional) Taints to created nodes to configure anti-affinity.
+  * `key` - (Required) A key must contain 1 to 63 characters starting with a letter or digit. Only letters, digits, hyphens (-), underscores (_), and periods (.) are allowed. A DNS subdomain name can be used as the prefix of a key.
+  * `value` - (Required) A value must start with a letter or digit and can contain a maximum of 63 characters, including letters, digits, hyphens (-), underscores (_), and periods (.).
+  * `effect` - (Optional) Available options are `NoSchedule`, `PreferNoSchedule`, and `NoExecute`.
+
 * `eip_ids` - (Optional) List of existing elastic IP IDs.
 
 -> If the `eip_ids` parameter is configured, you do not need to configure the `eip_count` and `bandwidth` parameters:
--- a/opentelekomcloud/services/cce/resource_opentelekomcloud_cce_node_v3.go
+++ b/opentelekomcloud/services/cce/resource_opentelekomcloud_cce_node_v3.go
@@ -281,6 +281,36 @@ func ResourceCCENodeV3() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"taints": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"key": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringMatch(clusterPoolTaintRegex, "Invalid key. "+
+								"Cluster pool taint key is 1 to 63 characters starting with a letter or digit. "+
+								"Only lowercase letters, digits, and hyphens (-) are allowed."),
+						},
+						"value": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringMatch(clusterPoolTaintRegex, "Invalid value. "+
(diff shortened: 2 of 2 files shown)


docs/resources/cce_node_v3.md
```diff
@@ -65,6 +65,11 @@ The following arguments are supported:
 
 * `annotations` - (Optional) Node annotation, key/value pair format. Changing this parameter will create a new resource.
 
+* `taints` - (Optional) Taints to created nodes to configure anti-affinity.
+  * `key` - (Required) A key must contain 1 to 63 characters starting with a letter or digit. Only letters, digits, hyphens (-), underscores (_), and periods (.) are allowed. A DNS subdomain name can be used as the prefix of a key.
+  * `value` - (Required) A value must start with a letter or digit and can contain a maximum of 63 characters, including letters, digits, hyphens (-), underscores (_), and periods (.).
```
outcatcher, line 70: Please emphasize values with `backquotes`

outcatcher (dismissed):
We will run acceptance tests ourselves :)

opentelekomcloud/acceptance/cce/resource_opentelekomcloud_cce_node_v3_test.go
```diff
@@ -249,6 +249,27 @@ func TestAccCCENodesV3EncryptedVolume(t *testing.T) {
...
+		ProviderFactories: common.TestAccProviderFactories,
+		CheckDestroy:      testAccCheckCCENodeV3Destroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCCENodeV3Taints,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCCENodeV3Exists(resourceNameNode, "opentelekomcloud_cce_cluster_v3.cluster_1", &node),
+					resource.TestCheckResourceAttr(resourceNameNode, "taints[0].key", "dedicated"),
+					resource.TestCheckResourceAttr(resourceNameNode, "taints[0].value", "database"),
+					resource.TestCheckResourceAttr(resourceNameNode, "taints[0].effect", "NoSchedule"),
```
lego963: can you change `taints[0].*` to `taints.0.*`?

lego963 (changes requested):

docs/resources/cce_node_v3.md
0xd3e: Can you elaborate on that please? What do you mean, when you are referring to values? The arguments (as Terraform calls them) like `taints`, `key` and `value` are already enclosed in backticks/backquotes.

The docs were copied from cce_node_pool_v3, so if something is wrong here it is also wrong there.

outcatcher: I was meaning quoting also the following:
```
A key must contain `1` to `63` characters...
Only letters, digits, hyphens (`-`), underscores (`_`), and periods (`.`) are allowed.
```
I suppose we don't always do like that ourselves :) 

P.S.
Actually, it's done only in 4 resources against 33 without backquoting such values. For CCE it's not done in other resources docs. So it's okay as it's now.

opentelekomcloud/acceptance/cce/resource_opentelekomcloud_cce_node_v3_test.go
0xd3e: This should be fixed now

outcatcher:
Okay, we got acceptance tests failing:
```
=== RUN   TestAccCCENodesV3Taints
    resource_opentelekomcloud_cce_node_v3_test.go:255: Step 1/1 error: Check failed: Check 2/4 error: opentelekomcloud_cce_node_v3.node_1: Attribute 'taints.0.key' not found
--- FAIL: TestAccCCENodesV3Taints (661.58s)

FAIL

Process finished with the exit code 1

```

For some reason `Taints` are empty in `nodes.Get` result. At the same time, via `kubectl` taints are set. We will investigate further.

---

### UPDATE
they are just missing in the response
image

outcatcher:
@0xIDANT As a workaround, you can skip taints reading for now (remove `d.Set("taints", node.Spec.Taints),`)

outcatcher:
Internal incident created: https://jira.tsi-dev.otc-service.com/browse/OCS-4091. But no guarantees it will be fixed in the nearest future.

0xd3e:
@outcatcher I applied your suggested changes and removed the acceptance test for the taints, since it wouldn't make any sense when the taints are missing.

lego963 (approved):

lego963:
@0xIDANT thanks for contribution :)

Issue #1186, opened 2021-07-01
WAF precise protection rule invalid logic for ip category


pschmidbauer:
# Terraform Version
1.0.0

# Affected Resource(s)
opentelekomcloud_waf_preciseprotection_rule_v1

# Debug Output/Panic Output
Terraform debug output https://gist.github.com/pschmidbauer/f9494164f23c61ab771e5924afd57388

# Steps to Reproduce
1. Add resource and try to set logic to "not_equal" for IP
`resource "opentelekomcloud_waf_preciseprotection_rule_v1" "only_allow_office_ip" {
  policy_id = opentelekomcloud_waf_policy_v1.policy_1.id
  name      = "only-office-ip"
  conditions {
    category = "ip"
    contents = ["XXX.XX.XXX.XX"]
    logic    = 4
  }

  action_category = "block"
  priority        = 4
}`
2. terraform apply

# Expected Behavior
A precise protection rule is created and all IP addresses are blocked except for the provided IP which can be an office IP for example. As the documentation says you can use 3 or 4 value for category ip

# Actual Behavior
It says the logic is invalid as you can see in the above provided GitHub gist

# Important Factoids
OTC provider version: 1.24.2

# References
None

outcatcher:
Fixed in #1193

PR #1193, opened 2021-07-06
[WAF] precise protection rule change logic parameter to string

cgroschupp:
the logic parameter is a string and no longer an int.

BREAKING CHANGE
It is possible to make it compatible, but I need feedback if that is necessary.

**Remove the `replace` in go.mod before merging.**
## PR Checklist

* [x] Refers to: #1186
* [x] Documentation updated.
* [x] Schema updated.

### DIFF
--- a/opentelekomcloud/services/waf/resource_opentelekomcloud_waf_preciseprotection_rule_v1.go
+++ b/opentelekomcloud/services/waf/resource_opentelekomcloud_waf_preciseprotection_rule_v1.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	golangsdk "github.com/opentelekomcloud/gophertelekomcloud"
 	"github.com/opentelekomcloud/gophertelekomcloud/openstack/waf/v1/preciseprotection_rules"
 
@@ -64,9 +65,10 @@ func ResourceWafPreciseProtectionRuleV1() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"category": {
-							Type:     schema.TypeString,
-							Required: true,
-							ForceNew: true,
+							Type:         schema.TypeString,
+							Required:     true,
+							ForceNew:     true,
+							ValidateFunc: validation.StringInSlice([]string{"path", "user-agent", "ip", "params", "cookie", "referer", "header"}, false),
 						},
 						"index": {
 							Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


opentelekomcloud/services/waf/resource_opentelekomcloud_waf_preciseprotection_rule_v1.go
```diff
@@ -74,9 +75,12 @@ func ResourceWafPreciseProtectionRuleV1() *schema.Resource {
 							ForceNew: true,
 						},
 						"logic": {
-							Type:     schema.TypeInt,
+							Type:     schema.TypeString,
 							Required: true,
 							ForceNew: true,
+							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+								return common.ValidateStringList(v, k, []string{"contain", "not_contain", "equal", "not_equal", "prefix", "not_prefix", "suffix", "not_suffix"})
+							},
```
outcatcher: There is `validation.StringInSlice` for such cases, please use it.

outcatcher (changes requested):
@cgroschupp 

I suppose it would be better to make a breaking change in the next release, we can increase the minor version (to `v1.25.0`) and schedule it to 14.07. But if it wasn't working, it's just a bug fix, not a breaking change.

Also, I would suggest updating `go.mod` after https://github.com/opentelekomcloud/gophertelekomcloud/pull/206 is merged. 

We can also make a same-day-release of 1.25.0 with backward compatibility breaking in case this topic is really urgent.

cgroschupp:
> @cgroschupp
> 
> I suppose it would be better to make a breaking change in the next release, we can increase the minor version (to `v1.25.0`) and schedule it to 14.07. But if it wasn't working, it's just a bug fix, not a breaking change.
> 
> Also, I would suggest updating `go.mod` after [opentelekomcloud/gophertelekomcloud#206](https://github.com/opentelekomcloud/gophertelekomcloud/pull/206) is merged.
> 
> We can also make a same-day-release of 1.25.0 with backward compatibility breaking in case this topic is really urgent.

I think it had never worked, but I'm not sure. The logic type was wrong and the category in the acceptance test was wrong.

outcatcher:
> > @cgroschupp
> > I suppose it would be better to make a breaking change in the next release, we can increase the minor version (to `v1.25.0`) and schedule it to 14.07. But if it wasn't working, it's just a bug fix, not a breaking change.
> > Also, I would suggest updating `go.mod` after [opentelekomcloud/gophertelekomcloud#206](https://github.com/opentelekomcloud/gophertelekomcloud/pull/206) is merged.
> > We can also make a same-day-release of 1.25.0 with backward compatibility breaking in case this topic is really urgent.
> 
> I think it had never worked, but I'm not sure. The logic type was wrong and the category in the acceptance test was wrong.

I suppose if there was no issue before, it was not used much. We can consider this just a fix without much risk.

outcatcher:
Let's wait for opentelekomcloud/gophertelekomcloud#206 to be merged to get rid of `replace` in dependencies :)

cgroschupp:
@outcatcher done

lego963:
@cgroschupp can you run acceptance test?

lego963 (approved):

cgroschupp:
@lego963 
```sh
TF_ACC=1 go test -run TestAccWafPreciseProtectionRuleV1_basic -v -timeout 720m ./opentelekomcloud/acceptance/waf/...
=== RUN   TestAccWafPreciseProtectionRuleV1_basic
--- PASS: TestAccWafPreciseProtectionRuleV1_basic (11.71s)
PASS
ok  	github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/waf	12.438s
```

outcatcher (approved):

Issue #1256, opened 2021-08-06
Fix lint errors in the provider


outcatcher:
# Description of the enhancement
Currently `golangci-lint` is made non-voting because of a lot of lint errors. Those should be fixed

PR #1259, opened 2021-08-06
Fix lint for `common` package

outcatcher:
Fix lint for `common` package

## PR Checklist

* [x] Refers to: #1256

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,3 +29,8 @@ linters:
     - vet
     - vetshadow
     - whitespace
+
+linters-settings:
+  misspell:
+    ignore-words:
+      - unknwon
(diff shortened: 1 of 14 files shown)


.golangci.yml
```diff
@@ -29,3 +29,8 @@ linters:
     - vet
     - vetshadow
     - whitespace
+
+linters-settings:
+  misspell:
+    ignore-words:
+      - unknwon
```
anton-sidelnikov, line 36: unknown?

anton-sidelnikov (changes requested):

.golangci.yml
outcatcher: There is a library `"github.com/unknwon/com"` and spellcheck think this is misspell :)

anton-sidelnikov (approved):

PR #1890, opened 2022-08-19
[CES]: module change after refactoring + deprecated field removal

artem-lifshits:
Links for API calls updated, deprecated field removed.
For additional info check: https://docs.otc.t-systems.com/api/ces/ces_03_0031.html

## PR Checklist
* [x] Tests updated.
* [x] Documentation updated.
* [x] Schema updated.
* [x] Release notes added.

## Acceptance Steps Performed

```
=== RUN   TestCESAlarmRule_basic
=== PAUSE TestCESAlarmRule_basic
=== CONT  TestCESAlarmRule_basic
--- PASS: TestCESAlarmRule_basic (113.57s)
PASS

```

### DIFF
--- a/opentelekomcloud/acceptance/ces/resource_opentelekomcloud_ces_alarmrule_test.go
+++ b/opentelekomcloud/acceptance/ces/resource_opentelekomcloud_ces_alarmrule_test.go
@@ -58,7 +58,7 @@ func TestAccCheckCESV1AlarmValidation(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config:      testCESAlarmRuleValidation,
-				ExpectError: regexp.MustCompile("Error: either `alarm_actions` or `ok_actions` should be specified+"),
+				ExpectError: regexp.MustCompile("Error: either `alarm_actions` or `ok_actions` should be specified.+"),
 			},
 		},
 	})
--- a/opentelekomcloud/acceptance/ces/resource_opentelekomcloud_ces_alarmrule_test.go
+++ b/opentelekomcloud/acceptance/ces/resource_opentelekomcloud_ces_alarmrule_test.go
@@ -2,6 +2,7 @@ package acceptance
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
@@ -45,6 +46,19 @@ func TestCESAlarmRule_basic(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(resourceAlarmRuleName, "alarm_enabled", "false"),
 				),
+			},
+		},
+	})
+}
+
+func TestAccCheckCESV1AlarmValidation(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { common.TestAccPreCheck(t) },
+		ProviderFactories: common.TestAccProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      testCESAlarmRuleValidation,
+				ExpectError: regexp.MustCompile("Error: either `alarm_actions` or `ok_actions` should be specified+"),
 			},
 		},
 	})
@@ -194,5 +208,46 @@ resource "opentelekomcloud_ces_alarmrule" "alarmrule_1" {
       opentelekomcloud_smn_topic_v2.topic_1.topic_urn
     ]
   }
+}
+`, common.DataSourceSubnet)
+
+var testCESAlarmRuleValidation = fmt.Sprintf(`
+%s
+
+resource "opentelekomcloud_compute_instance_v2" "vm_1" {
+  name        = "instance_1"
+  image_name  = "Standard_Debian_11_latest"
+  flavor_name = "s3.large.2"
+
+  network {
+    uuid = data.opentelekomcloud_vpc_subnet_v1.shared_subnet.network_id
+  }
+}
+
+resource "opentelekomcloud_smn_topic_v2" "topic_1" {
+  name         = "topic_1"
+  display_name = "The display name of topic_1"
+}
+
+resource "opentelekomcloud_ces_alarmrule" "alarmrule_1" {
+  alarm_name = "alarm_rule1"
+
+  metric {
+    namespace   = "SYS.ECS"
+    metric_name = "network_outgoing_bytes_rate_inband"
+    dimensions {
+      name  = "instance_id"
+      value = opentelekomcloud_compute_instance_v2.vm_1.id
+    }
+  }
+  condition {
+    period              = 300
+    filter              = "average"
+    comparison_operator = ">"
+    value               = 6
+    unit                = "B/s"
+    count               = 1
+  }
+  alarm_action_enabled = true
 }
 `, common.DataSourceSubnet)
--- a/opentelekomcloud/acceptance/ces/resource_opentelekomcloud_ces_alarmrule_test.go
+++ b/opentelekomcloud/acceptance/ces/resource_opentelekomcloud_ces_alarmrule_test.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
-	"github.com/opentelekomcloud/gophertelekomcloud/openstack/cloudeyeservice/alarmrule"
+	"github.com/opentelekomcloud/gophertelekomcloud/openstack/ces/v1/alarms"
 	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/common/quotas"
 	ecs "github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/ecs"
 
@@ -18,7 +18,7 @@ import (
 const resourceAlarmRuleName = "opentelekomcloud_ces_alarmrule.alarmrule_1"
 
 func TestCESAlarmRule_basic(t *testing.T) {
-	var ar alarmrule.AlarmRule
+	var ar alarms.MetricAlarms
 
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck: func() {
@@ -63,7 +63,7 @@ func testCESAlarmRuleDestroy(s *terraform.State) error {
 		}
 
 		id := rs.Primary.ID
-		_, err := alarmrule.Get(client, id).Extract()
+		_, err := alarms.ShowAlarm(client, id)
 		if err == nil {
 			return fmt.Errorf("alarm rule still exists")
 		}
@@ -72,7 +72,7 @@ func testCESAlarmRuleDestroy(s *terraform.State) error {
 	return nil
 }
 
-func testCESAlarmRuleExists(n string, ar *alarmrule.AlarmRule) resource.TestCheckFunc {
+func testCESAlarmRuleExists(n string, ar *alarms.MetricAlarms) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		rs, ok := s.RootModule().Resources[n]
 		if !ok {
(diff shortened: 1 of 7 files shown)


anton-sidelnikov (dismissed):

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
```diff
@@ -256,10 +232,11 @@ func ResourceAlarmRule() *schema.Resource {
...
 			"alarm_action_enabled": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				Default:  true,
-				ForceNew: true,
+				Type:         schema.TypeBool,
+				Optional:     true,
+				Default:      false,
+				ForceNew:     true,
+				AtLeastOneOf: []string{"alarm_actions", "ok_actions"},
```
outcatcher: That's not a correct usage of the `AtLeastOneOf`. It refers to at least one of attributes in the list (or this attribute) to be set.

```
	// AtLeastOneOf is a set of attribute paths, including this attribute,
	// in which at least one of the attributes must be configured. This
	// implements the validation logic declaratively within the schema and can
	// trigger earlier in Terraform operations, rather than using create or
	// update logic which only triggers during apply.
	//
	// Only absolute attribute paths, ones starting with top level attribute
	// names, are supported. Attribute paths cannot be accurately declared
	// for TypeList (if MaxItems is greater than 1), TypeMap, or TypeSet
	// attributes. To reference an attribute under a single configuration block
	// (TypeList with Elem of *Resource and MaxItems of 1), the syntax is
	// "parent_block_name.0.child_attribute_name".
	AtLeastOneOf []string
```

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
```diff
@@ -326,37 +303,39 @@ func resourceAlarmRuleCreate(ctx context.Context, d *schema.ResourceData, meta i
 		return fmterr.Errorf(errCreationClient, err)
 	}
 
+	alarmEnabled := d.Get("alarm_enabled").(bool)
+	alarmActionEnabled := d.Get("alarm_action_enabled").(bool)
+
 	metric := getMetricOpts(d)
 	conditionListRaw := d.Get("condition").([]interface{})
 	conditionElement := conditionListRaw[0].(map[string]interface{})
```
outcatcher, line 317: `conditionListRaw` length should be checked first to avoid panic

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
```diff
@@ -371,14 +350,16 @@ func resourceAlarmRuleRead(ctx context.Context, d *schema.ResourceData, meta int
 		return fmterr.Errorf(errCreationClient, err)
 	}
 
-	r, err := alarmrule.Get(client, d.Id()).Extract()
+	r, err := alarms.ShowAlarm(client, d.Id())
```
outcatcher: `r` is not a very good name for the variable.

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
```diff
@@ -371,14 +350,16 @@ func resourceAlarmRuleRead(ctx context.Context, d *schema.ResourceData, meta int
...
 	if err != nil {
 		return common.CheckDeletedDiag(d, err, "alarmrule")
 	}
-	log.Printf("[DEBUG] Retrieved alarm rule %s: %#v", d.Id(), r)
+	alarm := r[0]
```
outcatcher: Slice length should be checked first

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
outcatcher: SDK: Also, strange that `showAlarm` which is accepting unique ID returns list.

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
```diff
@@ -387,55 +368,38 @@ func resourceAlarmRuleRead(ctx context.Context, d *schema.ResourceData, meta int
...
-			"type":              insufficientActionItem.Type,
-			"notification_list": notificationList,
-		}
-		insufficientActionsInfo[i] = insufficientAction
-	}
-
-	okActionsInfo := make([]map[string]interface{}, len(r.OkActions))
-	for i, okActionItem := range r.OkActions {
+	okActionsInfo := make([]map[string]interface{}, len(alarm.OkActions))
+	for i, okActionItem := range alarm.OkActions {
```
outcatcher, line 406: `i` is unused, should be replaced with `_`

outcatcher (changes requested):

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
artem-lifshits: Well that's problems with API itself... I will change the name

outcatcher: Anyway, that doesn't relate to this PR :)

Aloento: I will change this name.

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
artem-lifshits: But it's used a bit later, otherwise it wouldn't compile

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
artem-lifshits: I didn't change that part :) But I will add the check, thank you

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
artem-lifshits: Should it really be checked? The API always returns a list, otherwise it fails, here's the doc:
https://docs.otc.t-systems.com/api/ces/ces_03_0031.html
It's pretty bad and have only closing bracket but I think you get the point

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
artem-lifshits: I see, what would you suggest to implement a state like this?
"If you set alarm_action_enabled to true, you must specify either alarm_actions or ok_actions."

artem-lifshits: I guess `requiredWith` on `alarm_actions`/ `ok_actions` would fit here

artem-lifshits: Nope it will make `alarm_action_enabled` mandatory if `alarm_actions`/`ok_actions` are enabled

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
Aloento: Ok, I will not change this name but change the return value.
It returns a list, but it always has at most one value.

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
Aloento: where is the AtLeastOneOf from?

artem-lifshits: Schema struct

Aloento: Sorry I didn't find it... could you give me a link?

outcatcher: > I see, what would you suggest to implement a state like this? "If you set alarm_action_enabled to true, you must specify either alarm_actions or ok_actions."

This should be implemented inside either in `CustomizeDiff` schema attribute (`schema.CustomizeDiffFunc`), there are some examples of that, or in exact update/create operation.


> Sorry I didn't find it... could you give me a link?

You can check the schema documentation:
https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/schema.go

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
outcatcher: Yes, I just noticed it nearby you changes 😄. It is not a problem now, but can become one in case of some changes, so if you're changing nearby code, it's a good idea to fix such potential problems 🙂

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
outcatcher: @Aloento There is a question though - if there is no values, should it return `nil, nil`, which is regarded as bad practice (and even a error for some linters), or standard SDK 404 error. I would suggest returning slice in case API returning slice 🤷 If I remember correctly, there are such cases already in the SDK.

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
outcatcher: In this case it's working correctly, but checking slice before addressing it is a good practice 🙂 you can assume my comment a style recommendation, not a found problem.

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
outcatcher: My bad, haven't noticed

artem-lifshits:
=== RUN   TestCESAlarmRule_basic
=== PAUSE TestCESAlarmRule_basic
=== CONT  TestCESAlarmRule_basic
--- PASS: TestCESAlarmRule_basic (126.81s)
=== RUN   TestAccCheckCESV1AlarmRestrictions
--- PASS: TestAccCheckCESV1AlarmRestrictions (11.01s)
PASS

opentelekomcloud/services/ces/resource_opentelekomcloud_ces_alarmrule.go
artem-lifshits: updated

anton-sidelnikov (dismissed):

Aloento (dismissed):

opentelekomcloud/acceptance/ces/resource_opentelekomcloud_ces_alarmrule_test.go
```diff
@@ -50,6 +51,19 @@ func TestCESAlarmRule_basic(t *testing.T) {
...
 }
 
+func TestAccCheckCESV1AlarmValidation(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { common.TestAccPreCheck(t) },
+		ProviderFactories: common.TestAccProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      testCESAlarmRuleValidation,
+				ExpectError: regexp.MustCompile("Error: either `alarm_actions` or `ok_actions` should be specified+"),
```
outcatcher: I suppose, it's a bit not the regex you want to have :) 
https://regex101.com/r/52NA6v/1

`d+` in the end basically meaning one or more `d` characters.

outcatcher (approved):

opentelekomcloud/acceptance/ces/resource_opentelekomcloud_ces_alarmrule_test.go
artem-lifshits: Thank you for noticing :)

anton-sidelnikov (approved):

PR #2011, opened 2022-11-30
[VPN] psk documentation and validation updates in `resource/vpnaas_site_connection_v2`

jrod-97:
I've updated the description of vpnaas_site_connection_v2 to reflect the real world requirements that are present for the PSK after I ran into some issues trying "any string" that was not accepted.

## PR Checklist

* [ ] Refers to: #xxx
* [x] Tests added/passed.
* [x] Documentation updated.
* [x] Schema updated.
* [ ] Release notes added.

### DIFF
--- a/docs/resources/vpnaas_site_connection_v2.md
+++ b/docs/resources/vpnaas_site_connection_v2.md
@@ -73,7 +73,7 @@ The following arguments are supported:
 
 * `peer_address` - (Required) The peer gateway public IPv4 or IPv6 address or FQDN.
 
-* `psk` - (Required) The pre-shared key. A valid value is any string that is between 6 and 128 characters long and does not contain the following characters: `<>&*'"?`
+* `psk` - (Required) The pre-shared key. A PSK can contain 6 to 128 characters. Spaces and the following special characters are not allowed: `<>&?*'"`.
 
 * `initiator` - (Optional) A valid value is `response-only` or `bi-directional`.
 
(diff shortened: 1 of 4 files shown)


vladimirvshivkov (dismissed):
I suppose it is a good idea to make this clear.

anton-sidelnikov:
@jrod-97 According the documentation you partly right: "A PSK can contain 6 to 128 characters. Spaces and question marks (?) are not allowed in a PSK. The PSK cannot contain only asterisks (*)"

jrod-97:
I've based my description on the popup in the web covering that appears when trying to paste a psk with a question mark in it. It says: `Spaces and the following special characters are not allowed: <>&?*'"`
Then you need to update the docs of both to clarify. 

The error message that is delivered via the API could also be improved. I got: `The PSK \"?\" is invalid`. And I was left wondering why my nice PSK is transmitted as `?` and suspected a API error at first.

anton-sidelnikov:
@jrod-97 
Yes, thank you, I will forward this to backend team. Also little bit rephrased doc. Added validation.

anton-sidelnikov (approved):

jrod-97:
The regex to validate is a nice addition. Should improve UX even more.
Thanks for forwarding to the backend team!

artem-lifshits (approved):

Aloento (approved):

Issue #2562, opened 2024-06-22
 Adding route to existing route table using terraform provider


simfeiOTC:
My question is: how can I add routes to the existing routing table?

My scenario is: 
1) I create a VPC with subnets and routing tables in a module A
2) I create VPN connection in a module B

Now I need to add a route to the subnet created in module A to the VPN gateway created in module B.

How can I do this?

- In theory I can import the resource and then update some attributes. Ok, so when I do:
```
import {
  to = opentelekomcloud_vpc_route_table_v1.rtb
  id = local.private-route-table.id
}

resource "opentelekomcloud_vpc_route_table_v1" "rtb" {
  vpc_id = local.vpc.id
  name = local.private-route-table.name
  route {
    destination = "10.0.0.0/8"
    type        = "vpn"
    nexthop     = module.otc-vpn-site-to-site.otc-tunnel.id
  }
}
```
my old routes are removed.

To prevent this I need to somehow merge old routes with new ones. But...

- There is no route table data source. So there is no way to get the existing routes for a specific route table.
- In documentation (https://registry.terraform.io/providers/opentelekomcloud/opentelekomcloud/latest/docs/resources/vpc_route_table_v1) the `route` block is described as  
> (Optional, List) Specifies the route object list. The route object is documented below.
however the declaration is a block, so when I do:
```
resource "opentelekomcloud_vpc_route_table_v1" "rtb" {
  route = []
}
```
that should work if route is the list - I get an error:
```
Error: Unsupported argument

  on main.tf line 16, in resource "opentelekomcloud_vpc_route_table_v1" "rtb":
  16:   route = []

An argument named "route" is not expected here. Did you mean to define a block of type "route"?
```

As usual I googled, read the docs, but... I didn't manage to find how to do such a simple action as just in general updating routes in the existing route table. 

And I know for sure that this is very standard action, for example:
- I can create individual route object inside route table in AWS CloudFormation (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-route.html, https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route)
- I can create individual routes using GCP (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_route) and even though I can't get all the routes - I can import the individual one w/o affecting any other.
[...]

PR #2567, opened 2024-06-26
[VPC] Add data source for route tables collection

apankratov:
## PR Checklist

* [x] Refers to: #2562 
* [x] Tests added/passed.
* [x] Documentation updated.
* [x] Schema updated.
* [x] Release notes added.

## Acceptance Steps Performed

> [!CAUTION]
>
> - No acceptance tests were run. This is external PR, I just don't have OTC environment in which I can run them. However I created the test that _in theory_ should pass.
>
> - Release notes were not updated as I am not the owner of the repo, I can't do releases.

Hopefully this is just the first PR in the series to enhance VPC routing support by provider. And just in case -  sorry, golang is not my native language, but I tried.

### DIFF
--- a/docs/data-sources/vpc_route_table.md
+++ b/docs/data-sources/vpc_route_table.md
@@ -44,7 +44,7 @@ In addition to all arguments above, the following attributes are exported:
 * `subnets` - An array of one or more subnets associating with the route table.
 
 * `region` - The region in which belongs the vpc route table.
-*
+
 * `route` - The route object list. The [route object](#route_object) is documented below.
 
 <a name="route_object"></a>
--- a/docs/data-sources/vpc_route_tables_v1.md
+++ b/docs/data-sources/vpc_route_tables_v1.md
@@ -2,8 +2,6 @@
 subcategory: "Virtual Private Cloud (VPC)"
 ---
 
-#
-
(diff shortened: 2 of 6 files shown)


anton-sidelnikov:
already there: data "opentelekomcloud_vpc_route_table_v1" with more plain structure that fit to resource, but i will have a look later and check the tests

apankratov:
The structure you mention returns single RT, not a collection. Moreover, if more than 1 route table is returned by `List` call - the data source will return default route table.

So, if I have 2 route tables inside VPC (for example, one to route traffic to VPN tunnel and another one isolated w/o any non-local routes for databases) - I will have to carry name or id between modules.

While if I have a collection - I can get all route tables and then filter them.

Also using the resource you've mentioned you can't solve single task: knowing subnet ID get the associated route table. 

This collection resource solves _at least_ both of these.

anton-sidelnikov:
Oki @apankratov, so I fixed tests, and other minor issues, think we can merge it now.

anton-sidelnikov:
```
=== RUN   TestAccVpcRouteTablesV1DataSource_basic
=== PAUSE TestAccVpcRouteTablesV1DataSource_basic
=== CONT  TestAccVpcRouteTablesV1DataSource_basic
--- PASS: TestAccVpcRouteTablesV1DataSource_basic (82.30s)
PASS

```

anton-sidelnikov (approved):

Issue #3142, opened 2025-09-30
[IAM] Documentation for opentelekomcloud_identity_group_v3 is outdated


mattef92:
The documentation of the resource "opentelekomcloud_identity_group_v3" is outdated.
Path: https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/blob/devel/docs/resources/identity_group_v3.md

It's lacking some Attribute references:
- description
- id
- name

On accessing the resource in Terraform, there's also the region. Due to the fact it's not even stated in the API documentation, I'll leave this out for now.

Happy to provide a PR to get this updated.

mattef92:
After feedback in the PR, description and name are automatically exposed from the arguments section

PR #3144, opened 2025-09-30
[Doc]: Update IAM docs for groups

mattef92:
Adding missing attribute references to the documentation of the IAM groups

## PR Checklist

* [x] Refers to: #3142 
* [ ] Tests added/passed.
* [x] Documentation updated.
* [ ] Schema updated.
* [x] Release notes added.

### DIFF
--- a/docs/resources/identity_group_v3.md
+++ b/docs/resources/identity_group_v3.md
@@ -37,12 +37,9 @@ The following arguments are supported:
 
 ## Attributes Reference
 
-The following attributes are exported:
+In addition to all arguments above, the following attributes are exported:
 
-* `description` - See Argument Reference above.
-* `domain_id` - See Argument Reference above.
 * `id` - The resource ID
-* `name` - See Argument Reference above.
 
 ## Import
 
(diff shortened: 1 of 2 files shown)


docs/resources/identity_group_v3.md
```diff
@@ -39,7 +39,10 @@ The following arguments are supported:
 
 The following attributes are exported:
 
+* `description` - See Argument Reference above.
```
anton-sidelnikov: In fact, we don’t need to list all the attributes here. By default, all arguments are exported, so there’s no need to specify them explicitly. Additionally, id is a default attribute for every resource, which makes mentioning it optional. Do you really want to include it?

anton-sidelnikov (changes requested):

docs/resources/identity_group_v3.md
mattef92: Thanks for the insight, haven't known this. Checked other documentations and would leave this out for name and description. A question which came to my mind on your answer: why is domain_id listed in here? It's also an argument, so I would remove it instead based on your answer.

Regarding the ID I was irritated, that it wasn't stated in the attributes. When there is no ID e.g. in the official Helm provider, the import is also done via namespace/name.
The group import is done on the group id, which I haven't found in the documentation.

Maybe that's just a personal problem. I'll adjust the documentation for the attributes and used the phrasing from user creation. Liked that one due to mentioning additional attributes to the arguments. Let me know, if that fits you

anton-sidelnikov (approved):

PR #3177, opened 2025-10-20
[Fix] import opentelekomcloud_evs_volume_v3 (#3176)

### LINKED ISSUE #3176: Importing EVS volume wants to replace it.
# Terraform provider version

Current `devel`

# Affected Resource(s)

opentelekomcloud_evs_volume_v3

the existing volume has "Volume Sharing" (multiattach) disabled and is a VBD volume. It has been created with the not-recommended "opentelekomcloud_blockstorage_volume_v2" before, but that should make no difference.

# Terraform Configuration Files
```hcl
resource "opentelekomcloud_evs_volume_v3" "my_data_volume" {
  name = "my-data-volume"
  description = "my data volume"
  volume_type = "GPSSD"
  size = 60
  availability_zone = "eu-de-01"

  # tried these later, no effect:
  multiattach = false # is already default
  cascade = false # defaults to true, but I tried both.
  device_type = "VBD" # is already default
  lifecycle {
    prevent_destroy = true
    ignore_changes = [
    ]
  }
}
```

# Steps to Reproduce

1. Import resource (provide`import` block)
1. `tofu apply`

# Expected Behavior
Import the existing volume.

# Actual Behavior

It tries to replace the volume:

````
  # opentelekomcloud_evs_volume_v3.my_data_volume must be replaced
  # (imported from "3d4e786c-5c3b-4ebf-ac03-25cfb036c2cd")
  # Warning: this will destroy the imported resource
-/+ resource "opentelekomcloud_evs_volume_v3" "opentelekomcloud_evs_volume_v3" {
      ~ attachment        = [] -> (known after apply)
        availability_zone = "eu-de-01"
      + cascade           = false # forces replacement
      + description       = "my data volume"
      + device_type       = "VBD" # forces replacement
      ~ id                = "3d4e786c-5c3b-4ebf-ac03-25cfb036c2cd" -> (known after apply)
      + multiattach       = false # forces replacement
        name              = "my-data-volume"
        size              = 60
        volume_type       = "GPSSD"
      ~ wwn               = "6888603000007cf8fa16890d47402814" -> (known after apply)

      - timeouts {}
    }
````

Looks like it does not retrieve the actual values for these three settings, despite them being both the default and likely retrievable via API.

# Important Factoids

I can work around this with

```
  lifecycle {
    prevent_destroy = true
    ignore_changes = [
      multiattach,
      cascade,
      device_type
    ]
  }
```

but this is ugly, as these cannot be changed afterwards anyways.

Happy to provide a PR, but I have no idea what in the source file influences the import behaviour.

### PR DESCRIPTION
**Pull request by MartinEmrich**

Proposal to fix #3176 :

* "cascade" is a delete option, thus not applicable for "ForceNew".
* "multiattach" is now taken from platform.
* "device_type" is inferred from volume metadata "hw:passthrough"

## PR Checklist

* [x] Refers to: #3176
* [x] Tests added/passed.
* [ ] Documentation updated.
* [x] Schema updated.
* [ ] Release notes added.

## Acceptance Steps Performed
Lacking a (cost-incurring) test platform: None! Please test/review before merge.

### DIFF
--- a/opentelekomcloud/acceptance/evs/resource_opentelekomcloud_evs_volume_v3_test.go
+++ b/opentelekomcloud/acceptance/evs/resource_opentelekomcloud_evs_volume_v3_test.go
@@ -18,42 +18,72 @@ import (
 
 const resourceVolumeV3Name = "opentelekomcloud_evs_volume_v3.volume_1"
 
+func getVolumeResourceFunc(conf *cfg.Config, state *terraform.ResourceState) (interface{}, error) {
+	c, err := conf.BlockStorageV3Client(env.OS_REGION_NAME)
+	if err != nil {
+		return nil, fmt.Errorf("error creating OpenTelekomCloud BlockStorage V3 client: %s", err)
+	}
+	return volumes.Get(c, state.Primary.ID).Extract()
+}
+
 func TestAccEvsStorageV3Volume_basic(t *testing.T) {
 	var volume volumes.Volume
+
+	rc := common.InitResourceCheck(
+		resourceVolumeV3Name,
+		&volume,
+		getVolumeResourceFunc,
+	)
+
 	t.Parallel()
 	quotas.BookMany(t, volumeQuotas(12))
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:          func() { common.TestAccPreCheck(t) },
 		ProviderFactories: common.TestAccProviderFactories,
-		CheckDestroy:      testAccCheckEvsStorageV3VolumeDestroy,
+		CheckDestroy:      rc.CheckResourceDestroy(),
 		Steps: []resource.TestStep{
 			{
 				Config: testAccEvsStorageV3VolumeBasic,
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckEvsStorageV3VolumeExists(resourceVolumeV3Name, &volume),
+					rc.CheckResourceExists(),
 					resource.TestCheckResourceAttr(resourceVolumeV3Name, "name", "volume_1"),
 				),
 			},
 			{
 				Config: testAccEvsStorageV3VolumeUpdate,
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckEvsStorageV3VolumeExists(resourceVolumeV3Name, &volume),
+					rc.CheckResourceExists(),
 					resource.TestCheckResourceAttr(resourceVolumeV3Name, "name", "volume_1-updated"),
 				),
+			},
+			{
+				ResourceName:      resourceVolumeV3Name,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"cascade",
+				},
 			},
 		},
 	})
 }
 
 func TestAccEvsStorageV3Volume_tags(t *testing.T) {
+	var volume volumes.Volume
+
+	rc := common.InitResourceCheck(
+		resourceVolumeV3Name,
+		&volume,
+		getVolumeResourceFunc,
+	)
 	t.Parallel()
 	quotas.BookMany(t, volumeQuotas(12))
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:          func() { common.TestAccPreCheck(t) },
 		ProviderFactories: common.TestAccProviderFactories,
-		CheckDestroy:      testAccCheckEvsStorageV3VolumeDestroy,
+		CheckDestroy:      rc.CheckResourceDestroy(),
 		Steps: []resource.TestStep{
 			{
 				Config: testAccEvsStorageV3VolumeTags,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by anton-sidelnikov** [comment]
Hi @MartinEmrich here also you must put 'cascade'  to ignore:
```
  lifecycle {
    ignore_changes = [
      cascade,
    ]
  }
```
The ignore_changes block is standard behavior in Terraform, so it’s perfectly fine to use. Even in large cloud environments, you’ll often find it applied to many resources

**Comment by anton-sidelnikov** [comment]
@MartinEmrich Also, please give me access to push, i will push tests update

**Comment by MartinEmrich** [comment]
@anton-sidelnikov I added you with write permissions, please try.

On `ignore_changes`, IMHO that option is for stack authors to ignore changes on specific attributes based on a specific use case. It should not be necessary to get a resource to work in the first place. So making it non-"ForceNew" for an attribute that no really sits on the provider/server side, I think that's a cleaner approach.

**Comment by anton-sidelnikov** [comment]
No, it depends on a few factors. The main problem happens when the resource doesn’t set values during the read phase, the API doesn’t return some fields, or you’re using attributes like cascade. So even if you set the parameter as not ForceNew, Terraform still asks you to ignore it since it’s not stored in the state. Trying to “fix” this could cause more problems for other users.

ForceNew parameter is False by default if not set

**Review by anton-sidelnikov — DISMISSED** [review]

**Review by artem-lifshits — DISMISSED** [review]

**Review by anton-sidelnikov — APPROVED** [review]

**Review by artem-lifshits — APPROVED** [review]

**Comment by MartinEmrich** [comment]
> So even if you set the parameter as not ForceNew, Terraform still asks you to ignore it since it’s not stored in the state. Trying to “fix” this could cause more problems for other users.

That would be valid for attributes actually backed by the provider (for example, "disk size" or "availability zone").

But `cascade` is (to my knowledge) a setting not stored on OpenTelekomCloud anywhere, it is purely stored in the terraform state, and has no effect until the resource is deleted. Only then, it is used during the delete API call to delete any snapshots from the volume.

So when importing a resource, there is no "previous value" of `cascade`. And changing it does not change anything on OTC side, so  in a way it is modifiable at runtime without interruption, thus *ForceNew: false*.

I see no scenario where this would break anything:
- User did not care/did not know about  `cascade`: Default would apply all along, so silently applying the default to the imported resource would be the least suprising approach.
- User wanted it explicity `true` or `false`: They would have it written in their resource definition, so, too, it would be applied during import.

In both (or any) cases, "applying" just means writing it into the terraform state, no change to the infrastructure is necessary. So deleting and replacing the volume is not necessary.

In my case (the first), it now does what I would expect: apply the default `cascade` value to my imported volume's state without deleting and recreating it.

**Comment by anton-sidelnikov** [comment]
@MartinEmrich all default attributes are stored in the state, and Terraform compares the current and previous values each time. As a result, Terraform detects differences and displays them, even when the cascade parameter isn’t marked as ForceNew. This is expected behavior, it’s simply how Terraform works.

You can verify this yourself by building the provider binary from this branch, placing it alongside your infrastructure code, and adding a tfrc file with the following configuration to override the provider binary locally:
```
provider_installation {
    dev_overrides {
        "opentelekomcloud/opentelekomcloud" = "/your/local/path/to/provider/binary/"

    }
    direct {}
}

```

**Comment by MartinEmrich** [comment]
I am already using my locally-built provider from this branch for providing the pull request, and it works as I described and expected.

There is no previous state, as the object is being imported. During import, it _has_ to start fresh from scratch.

Example: first remove the object from the state:

```
 ./tofu state rm opentelekomcloud_evs_volume_v3.management_host_data_disk
```

Then, on the next plan, it will import it now, without trying to replace it: `tofu plan`:

```
...
  # opentelekomcloud_evs_volume_v3.management_host_data_disk will be updated in-place
  # (imported from "3d4e786c-5c3b-4ebf-ac03-25cfb036c2cd")
  ~ resource "opentelekomcloud_evs_volume_v3" "management_host_data_disk" {
        attachment        = [
            {
                device      = "/dev/vdb"
                id          = "3d4e786c-5c3b-4ebf-ac03-25cfb036c2cd"
                instance_id = "502e9d85-d7de-448f-ab92-3b136bdbda03"
            },
        ]
        availability_zone = "eu-de-01"
      + cascade           = true
        description       = "otc-teststack-managementhost data volume"
        device_type       = "VBD"
        id                = "3d4e786c-5c3b-4ebf-ac03-25cfb036c2cd"
        multiattach       = false
        name              = "otc-teststack-managementhost-datadisk"
        size              = 60
        volume_type       = "GPSSD"
        wwn               = "6888603000007cf8fa16890d47402814"

      - timeouts {}
    }
...
```

Exactly as expected: As the `ForceNew` attribute from `cascade` is no longer `true`, it does no longer try to replace the volume, but "updates" it in place during import. 

Note that it just lists `cascade` as new (little `+` symbol), as there could not have been any previous value for it during import.

Perfect!

PR #3271, opened 2026-02-17
docs: vpc_route_v2 hint how to find internal id

madmuffin1:
we recently needed go back to a tfstate backup, that was missing a route, so I had to figure out which ID to use to perform terraform import - and it took me quite a while.

not sure if notes like this are welcome, but it would have helped me greatly and saved me hours of trying and searching if it had been there.

## PR Checklist

* [ ] Refers to: #xxx
* [ ] Tests added/passed.
* [x] Documentation updated.
* [ ] Schema updated.
* [ ] Release notes added.

## Acceptance Steps Performed

```
=== RUN   TestAccSomethingV0_basic
--- PASS: TestAccSomethingV0_basic (101.71s)
=== RUN   TestAccSomethingV0_timeout
--- PASS: TestAccSomethingV0_timeout (128.67s)
PASS

Process finished with exit code 0
```

### DIFF
--- a/docs/resources/vpc_route_v2.md
+++ b/docs/resources/vpc_route_v2.md
@@ -62,3 +62,5 @@ VPC route can be imported using the `id`, e.g.
 ```sh
 terraform import opentelekomcloud_vpc_route_v2.vpc_route 2c7fs9f3-712b-18d1-940c-b50384177ee1
 ```
+
+-> NOTE: The internal `id` is not visible in the OTC UI. If you want to import an existing opentelekomcloud_vpc_route_v2, you can query the routes from `https://vpc.eu-de.otc.t-systems.com/v2.0/vpc/routes?vpc_id=<vpc_id>` to find the internal id. (see [API Reference: Querying VPC Routes](https://docs.otc.t-systems.com/virtual-private-cloud/api-ref/apis/vpc_route/querying_vpc_routes.html#vpc-route-0001) for more details on that API)


anton-sidelnikov:
Hi @madmuffin1, thank you for PR. I'm not against such notes it doc, but i prefer not to mention manual steps, in terraform you could use data-sources for obtain additional info, like:
```
data "opentelekomcloud_vpc_route_v2" "by_vpc_id" {
  vpc_id = opentelekomcloud_vpc_route_v2.route_1.vpc_id
}
```
i prefer to put this block in your note, and then we can merge it

madmuffin1:
@anton-sidelnikov tbh I do not understand how this helps. when importing a resource using terraform import I need to explicitly specify the ID on the terminal, using a data source is a different scenario imho.

anton-sidelnikov:
@madmuffin1, I prefer to stay within the Terraform environment because it avoids the need to use additional clients or tools to make these requests—Terraform already provides native ways to fetch data from the cloud without context switching. For import purposes, you can combine data sources with outputs to print the required IDs. Data sources also help store this information in state. I don't insist to completely change your note, but please also add part regarding data source

ChristianUlbrich:
What is the added value of this? I bet there are lot of other resources for which, there is not the internal id shown in the GUI. Furthermore this example does not pay attention to the actual OTC region. I am for consistency, I'd rather have this as a high level concept explained, than at a very specific resource.

The question would be, what is the actual thing to explain? - If you cannot find some OTC internals, that you do not know, then use the OTC API directly? Isn't that always the case with OTC (actually usually for me it is the other way around, I need to sniff on GUI interactions, to figure out proper params)?

Issue #3292, opened 2026-03-04
[CCE] opentelekomcloud_cce_node_pool_v3 preinstall and postinstall should forceNew


pvlkov:
# Terraform provider version
```hcl
provider "registry.terraform.io/opentelekomcloud/opentelekomcloud" {
  version     = "1.36.60"
  constraints = "1.36.60"
}
```

# Affected Resource(s)
`opentelekomcloud_cce_node_pool_v3`

# Output
```bash
# opentelekomcloud_cce_node_pool_v3.nodes["worker"] will be updated in-place
     ~ resource "opentelekomcloud_cce_node_pool_v3" "nodes" {
         id                       = "xxxx"
         name                     = "worker"
       ~ preinstall               = "xxxx" -> "xxxx"
         # (19 unchanged attributes hidden)
         # (2 unchanged blocks hidden)
     }
```

# Steps to Reproduce
1. Change "preinstall" or "postinstall" attribute in `opentelekomcloud_cce_node_pool_v3`
2. `terraform apply`

# Expected Behavior

According to documentation changing "preinstall" or "postinstall" attributes should create a new node_pool resource as documented https://registry.terraform.io/providers/opentelekomcloud/opentelekomcloud/latest/docs/resources/cce_node_pool_v3#preinstall-2

# Actual Behavior

The resource is updated in place and any changes have no effect until new nodes are provisioned in the pool.

PR #3293, opened 2026-03-04
Adds ForceNew to preinstall and postinstall fields of resource opentelekomcloud_cce_node_pool_v3

pvlkov:
Adds ForceNew to "preinstall" and "postinstall" fields of opentelekomcloud_cce_node_pool_v3 schema. As per documentation changes to these fields should force a recreation of the node_pool resource (see https://registry.terraform.io/providers/opentelekomcloud/opentelekomcloud/latest/docs/resources/cce_node_pool_v3#preinstall-2)

## PR Checklist

* [x] Refers to: #3292
* [ ] Tests added/passed.
* [x] Documentation updated. (Documentation is already reflecting the behavior after the change)
* [x] Schema updated.
* [ ] Release notes added.

# Tests performed
Ran `terraform apply` in our own CCE environment with a custom build of the terraform provider.
Output:
```
   # opentelekomcloud_cce_node_pool_v3.nodes["worker"] must be replaced
 -/+ resource "opentelekomcloud_cce_node_pool_v3" "nodes" {
       ...
       ~ preinstall               = "xxxx" -> "xxxx" # forces replacement
       ..
     }
```

Node pool was rotated as intended with the new preinstall changes applied. Did not test postinstall, but I don't see why that should behave any different.

### DIFF
--- a/opentelekomcloud/services/cce/resource_opentelekomcloud_cce_node_pool_v3.go
+++ b/opentelekomcloud/services/cce/resource_opentelekomcloud_cce_node_pool_v3.go
@@ -238,12 +238,14 @@ func ResourceCCENodePoolV3() *schema.Resource {
 				Optional:  true,
 				Computed:  true,
 				StateFunc: common.GetHashOrEmpty,
+				ForceNew:  true,
 			},
 			"postinstall": {
 				Type:      schema.TypeString,
 				Optional:  true,
 				Computed:  true,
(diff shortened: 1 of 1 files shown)


muneeb-jan (approved):

anton-sidelnikov:
Hi @pvlkov, if you really want to make this fields ForceNew, then need to cleanup update

Issue #3370, opened 2026-05-13
[VPC] `secondary_cidr` on `opentelekomcloud_vpc_v1` is non-functional and can only ever represent one CIDR


TobiWo:
## Terraform provider version

```bash
opentelekomcloud >= 1.36.x (all versions exposing `secondary_cidr` on vpc_v1)
```

## Affected Resource(s)

- opentelekomcloud_vpc_v1

# Terraform Configuration Files

```hcl
resource "opentelekomcloud_vpc_v1" "vpc" {
  name           = "tf-vpc"
  cidr           = "192.168.0.0/16"
  secondary_cidr = "23.9.0.0/16"
}
```

## Debug Output/Panic Output

No debug output required to demonstrate. The behavioural problems are
visible in the source.

## Steps to Reproduce

1. Create a VPC via Terraform with `secondary_cidr = "23.9.0.0/16"`.
2. Add a second secondary CIDR via the OTC console (the cloud allows up to 5).
3. Run `terraform plan`. The additional CIDR is silently masked, no drift.
4. Remove the secondary CIDR via the OTC console.
5. Run `terraform plan`. The resource still reports the old CIDR in state and
   does not reconcile to "removed".
6. Alternatively, change `secondary_cidr` in HCL: the update path indexes
   `vpcV3Get.SecondaryCidrs[0]` unconditionally, which panics if the cloud-side
   list is empty and removes the wrong CIDR if order differs from insertion.
7. Or remove the `secondary_cidr` line from HCL: the update path issues a
   destructive `RemoveSecondaryCidr` followed by an `AddSecondaryCidr("")`
   which fails. The CIDR is wiped from the cloud (or the apply errors out
   if subnets exist inside it).

## Expected Behavior

A user should be able to manage all secondary CIDR blocks on a VPC. Per the
OTC v3 API, up to 5 secondary CIDRs are supported per VPC. The provider should
reflect cloud-side state on `terraform plan` (drift detection) and avoid
unsafe indexing and destructive update sequences.

## Actual Behavior

Five distinct problems make the attribute unusable in practice:

**Bug 1 — Schema cannot represent the API.** `secondary_cidr` is
`TypeString`, so only one CIDR can ever be expressed. The cloud allows up
to 5.

**Bug 2 — Update path is unsafe.** `updateSecondaryCidr` in
`opentelekomcloud/services/vpc/resource_opentelekomcloud_vpc_v1.go`
[...]

PR #3371, opened 2026-05-13
[Feat] VPC: Add `opentelekomcloud_vpc_secondary_cidr_v3`

TobiWo:

Add a new resource `opentelekomcloud_vpc_secondary_cidr_v3` for managing up to 5 secondary CIDR blocks attached to an existing VPC via the v3 API endpoints (`PUT /v3/{project_id}/vpc/vpcs/{vpc_id}/{add,remove}-extend-cidr`).

The `secondary_cidr` attribute on `opentelekomcloud_vpc_v1` is not removed in this PR. Its graceful deprecation and removal will follow in a separate PR.

### New resource: `opentelekomcloud_vpc_secondary_cidr_v3`

- Schema: `vpc_id` (`Required + ForceNew`), `cidrs` (`TypeSet`, `MinItems=1`, `MaxItems=5`, IsCIDR), `region` (computed/optional). Resource ID is the parent VPC ID. This is exactly one resource per VPC.
- **Update** computes the set difference via `common.GetSetChanges` and issues at most one `RemoveSecondaryCidr` and one `AddSecondaryCidr` per change.
- **Read** uses `common.CheckDeletedDiag` for VPC-deleted, and overwrites `cidrs` from the live API response so external mutations are detected as drift.
- **Delete** issues a single `RemoveSecondaryCidr` with the full set; 404 is treated as already-gone.
- **Import** via `schema.ImportStatePassthroughContext` using the VPC ID.

### Related bug fixes (separate PRs)

Two bugs in `opentelekomcloud_vpc_v1` had to be fixed before this resource can coexist safely with a VPC managed by `vpc_v1`:

- **#3373**: Fixed the destructive update path: `secondary_cidr = "x" → ""` was removing the cloud-side CIDR and then calling `AddSecondaryCidr("")`, losing the CIDR and erroring mid-apply.
- **#3375**: Fixed `readSecondaryCidr` state pollution: The function was running unconditionally on every refresh and writing any CIDR it found into `secondary_cidr`, even when the user never set the field. CIDRs added by this new resource were silently claimed by `vpc_v1` state, triggering a removal on the next apply.

### Why a sub-resource and not a full `vpc_v3`

The v3 SDK has only `Get`, `List`, `AddSecondaryCidr`, `RemoveSecondaryCidr`. No `Create`, `Update`, or `Delete` functionality present. The OTC v3 API exposes no general PUT-on-VPC endpoint either; v1 still owns name/description/cidr updates. A full `vpc_v3` resource would add SDK code that calls v1 internally for half its operations and would invite users to manage one cloud entity through two parallel resources. The sub-resource approach mirrors the API shape (set-style add/remove against a parent VPC) and reuses the established provider pattern from `opentelekomcloud_vpc_bandwidth_associate_v2`.

## PR Checklist

- [x] Refers to: #3370
- [x] Tests added/passed.
- [x] Documentation updated.
- [ ] Schema updated. (N/A)
- [x] Release notes added.

## Acceptance Steps Performed

```text
=== RUN   TestAccVpcSecondaryCidrV3_basic
=== PAUSE TestAccVpcSecondaryCidrV3_basic
=== CONT  TestAccVpcSecondaryCidrV3_basic
--- PASS: TestAccVpcSecondaryCidrV3_basic (241.37s)
PASS
```

`TestAccVpcSecondaryCidrV3_basic` covers create, update (adding/removing CIDRs), and import in a single test suite using `common.InitResourceCheck`. Acceptance tests for `opentelekomcloud_vpc_v1` were updated and verified in #3373 and #3375.

### DIFF
--- a/opentelekomcloud/acceptance/vpc/resource_opentelekomcloud_vpc_secondary_cidr_v3_test.go
+++ b/opentelekomcloud/acceptance/vpc/resource_opentelekomcloud_vpc_secondary_cidr_v3_test.go
@@ -1,0 +1,158 @@
+package acceptance
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/opentelekomcloud/gophertelekomcloud/acceptance/tools"
+	VpcV3 "github.com/opentelekomcloud/gophertelekomcloud/openstack/vpc/v3/vpcs"
+
+	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/common"
+	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/common/quotas"
+	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/env"
+	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/common/cfg"
+)
+
+const resourceVpcSecondaryCidrV3Name = "opentelekomcloud_vpc_secondary_cidr_v3.test"
+
+func getVpcSecondaryCidrFunc(c *cfg.Config, state *terraform.ResourceState) (interface{}, error) {
+	client, err := c.NetworkingV3Client(env.OS_REGION_NAME)
+	if err != nil {
+		return nil, fmt.Errorf("error creating OpenTelekomCloud NetworkingV3 client: %w", err)
+	}
+	vpc, err := VpcV3.Get(client, state.Primary.ID)
+	if err != nil {
+		return nil, err
+	}
+	if len(vpc.SecondaryCidrs) == 0 {
+		return nil, fmt.Errorf("vpc %s has no secondary CIDRs", state.Primary.ID)
+	}
+	return vpc, nil
+}
+
+func TestAccVpcSecondaryCidrV3_basic(t *testing.T) {
+	t.Parallel()
+	quotas.BookOne(t, quotas.Router)
+	vpcName := tools.RandomString("tf-acc-sec-cidr-", 5)
+
+	var vpc VpcV3.Vpc
+	rc := common.InitResourceCheck(resourceVpcSecondaryCidrV3Name, &vpc, getVpcSecondaryCidrFunc)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { common.TestAccPreCheck(t) },
+		ProviderFactories: common.TestAccProviderFactories,
+		CheckDestroy:      rc.CheckResourceDestroy(),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccVpcSecondaryCidrV3OneCidr(vpcName),
+				Check: resource.ComposeTestCheckFunc(
+					rc.CheckResourceExists(),
+					resource.TestCheckResourceAttr(resourceVpcSecondaryCidrV3Name, "cidrs.#", "1"),
+					resource.TestCheckTypeSetElemAttr(resourceVpcSecondaryCidrV3Name, "cidrs.*", "23.9.0.0/16"),
+				),
+			},
+			{
+				Config: testAccVpcSecondaryCidrV3ThreeCidrs(vpcName),
+				Check: resource.ComposeTestCheckFunc(
+					rc.CheckResourceExists(),
+					resource.TestCheckResourceAttr(resourceVpcSecondaryCidrV3Name, "cidrs.#", "3"),
+					resource.TestCheckTypeSetElemAttr(resourceVpcSecondaryCidrV3Name, "cidrs.*", "23.9.0.0/16"),
+					resource.TestCheckTypeSetElemAttr(resourceVpcSecondaryCidrV3Name, "cidrs.*", "23.10.0.0/16"),
+					resource.TestCheckTypeSetElemAttr(resourceVpcSecondaryCidrV3Name, "cidrs.*", "23.11.0.0/16"),
+				),
+			},
+			{
+				Config: testAccVpcSecondaryCidrV3TwoCidrs(vpcName),
+				Check: resource.ComposeTestCheckFunc(
+					rc.CheckResourceExists(),
+					resource.TestCheckResourceAttr(resourceVpcSecondaryCidrV3Name, "cidrs.#", "2"),
+					resource.TestCheckTypeSetElemAttr(resourceVpcSecondaryCidrV3Name, "cidrs.*", "23.9.0.0/16"),
+					resource.TestCheckTypeSetElemAttr(resourceVpcSecondaryCidrV3Name, "cidrs.*", "23.11.0.0/16"),
+				),
+			},
+			{
+				Config: testAccVpcSecondaryCidrV3MaxCidrs(vpcName),
+				Check: resource.ComposeTestCheckFunc(
+					rc.CheckResourceExists(),
+					resource.TestCheckResourceAttr(resourceVpcSecondaryCidrV3Name, "cidrs.#", "5"),
+				),
+			},
+			{
+				ResourceName:      resourceVpcSecondaryCidrV3Name,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func testAccVpcSecondaryCidrV3OneCidr(vpcName string) string {
+	return fmt.Sprintf(`
+resource "opentelekomcloud_vpc_v1" "test" {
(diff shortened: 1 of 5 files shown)


TobiWo:
Hi @anton-sidelnikov @artem-lifshits, this PR needs a version-bump decision before it leaves draft. It contains a breaking change on opentelekomcloud_vpc_v1 (the secondary_cidr attribute is removed and a state schema upgrader runs on first plan), so I'd recommend a minor bump to `1.37.0` rather than continuing on `1.36.x`. The version is referenced once in vpc_secondary_cidr_v3.md in the downgrade note at the bottom. Once you confirm the target version I'll update that reference and move the PR out of draft.

anton-sidelnikov:
Hi @TobiWo, i will look soon on the code, i like your idea, maybe with some additions we will merge it soon

anton-sidelnikov:
@TobiWo, Good job, and I like the idea of introducing a separate opentelekomcloud_vpc_secondary_cidr_v3 resource.

However, I see a few problems for users who already manage one secondary CIDR via opentelekomcloud_vpc_v1.secondary_cidr. With the current approach, we would force them to migrate/change their resource immediately. I would prefer to handle this more gracefully.

My suggestion is:

1. Keep vpc_v1 unchanged in this PR.
2. Merge only the new opentelekomcloud_vpc_secondary_cidr_v3 resource for now.
3. I will fix the existing secondary CIDR removal behavior in vpc_v1 separately.
4. After that, we can deprecate secondary_cidr in vpc_v1 and guide users toward the new resource.

So please remove all changes related to vpc_v1 from this PR. Let’s merge only the new resource first

opentelekomcloud/acceptance/vpc/import_opentelekomcloud_vpc_secondary_cidr_v3_test.go
```diff
@@ -0,0 +1,33 @@
...
+		CheckDestroy:      testAccCheckVpcSecondaryCidrV3Destroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccVpcSecondaryCidrV3OneCidr(vpcName),
+			},
+			{
+				ResourceName:      resourceVpcSecondaryCidrV3Name,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
anton-sidelnikov: better not split but add this step to basic test

opentelekomcloud/acceptance/vpc/resource_opentelekomcloud_vpc_secondary_cidr_v3_test.go
```diff
@@ -0,0 +1,188 @@
...
+			// VPC itself is gone, which means secondary CIDRs are gone too.
+			return nil
+		}
+		if len(vpcInfo.SecondaryCidrs) > 0 {
+			return fmt.Errorf("vpc %s still has secondary CIDRs: %v", rs.Primary.ID, vpcInfo.SecondaryCidrs)
+		}
+	}
+
+	return nil
+}
```
anton-sidelnikov: please use resource_check.go from common

anton-sidelnikov: you can also use it for `exists` checking

opentelekomcloud/services/vpc/resource_opentelekomcloud_vpc_secondary_cidr_v3.go
```diff
@@ -0,0 +1,170 @@
...
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Computed: true,
+			},
```
anton-sidelnikov, line 33: i prefer not to use region as parameter at all, because we alway took region from provider config, you can left it as computed if you want to keep it in state

opentelekomcloud/services/vpc/resource_opentelekomcloud_vpc_secondary_cidr_v3.go
```diff
@@ -0,0 +1,170 @@
...
+			return nil
+		}
+		return fmterr.Errorf("error removing secondary CIDRs from VPC %s: %w", d.Id(), err)
+	}
+
+	return nil
+}
+
+// expandCidrSet converts a schema.Set of strings into a []string preserving the set's contents.
+func expandCidrSet(set *schema.Set) []string {
```
anton-sidelnikov: probably you can use common.ExpandToStringListBySet instead, please check

anton-sidelnikov (changes requested):

opentelekomcloud/acceptance/vpc/import_opentelekomcloud_vpc_secondary_cidr_v3_test.go
anton-sidelnikov: please look on refactored tests, for example in apigw

TobiWo:
Hi @anton-sidelnikov, all comments are implemented. However, the test for the new resource is hitting an issue rooted in `vpc_v1`, not in this PR. I'm flagging it because it affects how we proceed.

**Symptom:** `TestAccVpcSecondaryCidrV3_basic` fails on step 1 with a non-empty plan after apply:

```text
# opentelekomcloud_vpc_v1.test will be updated in-place
~ resource "opentelekomcloud_vpc_v1" "test" {
      id             = "..."
      name           = "tf-acc-sec-cidr-..."
    - secondary_cidr = "23.9.0.0/16" -> null
      tags           = {}
  }
```

**Root cause:** `readSecondaryCidr` in `resource_opentelekomcloud_vpc_v1.go` always calls `VpcV3.Get` and writes whatever it finds into the `secondary_cidr` attribute, regardless of whether the user actually set the field in HCL. When my new resource adds `23.9.0.0/16` to the same VPC's `extend_cidrs`, the next refresh of `vpc_v1.test` claims that CIDR as its own state. Plan then proposes "remove" (and on apply would actually remove it), fighting the new resource.

Your `#3373` correctly addressed the destructive `secondary_cidr = "x" → ""` path on Update, but it left `readSecondaryCidr` untouched. So this Read-side state pollution survives.

**Heads-up for transparency:** I remember that this was one of the bugs that motivated the original StateUpgrader-based removal in this PR. The Read-side pollution is one of five issues on the field that I documented in the original PR description. I fully accept the call to do this incrementally but it does mean that until `readSecondaryCidr` is fixed, every user combining the new resource with a VPC managed via `vpc_v1` will hit this on `terraform plan`. Worth knowing as you plan the deprecation timeline.

**Minimal fix in `vpc_v1`:** an early-return guard so the v3 API is only consulted when the user actually set the field:

```go
func readSecondaryCidr(d *schema.ResourceData, config *cfg.Config) error {
    if d.Get("secondary_cidr").(string) == "" {
        return nil
    }
    // ... existing logic ...
}
```

Six lines, no behaviour change for users who do set the field, removes the conflict for users who combine `vpc_v1` with the new resource.

**How would you like to proceed?**

1. You land a small follow-up fix on `devel` (could be the early-return above or a different shape you prefer), then I rebase this PR over it.
2. I open a separate PR with just the `readSecondaryCidr` fix; once it merges I rebase this one.
3. I add `ExpectNonEmptyPlan: true` to the affected step here as a workaround. The test passes but ships with a known bug that surfaces for any user combining the two resources.

I'd lean toward 1 or 2. The workaround in 3 hides the real issue and the user impact persists until the v1 attribute is fully deprecated. Happy to do whichever you prefer.

Edit:
**FYI:** I kept my `vpc_v1` changes on `feature/vpc-v1-secondary-cidr-cleanup`.

anton-sidelnikov:
Hi @TobiWo, yes, I agree with you. I just want to fix everything step by step, and we only need to make sure it’s all done before this week’s release.

Right now, the main thing we need to fix is the read flow. I’m completely fine with either option 1 or 2. If you want, feel free to implement it, otherwise, I’ll handle it soon. After that, we can rebase your new resource and merge everything into devel

TobiWo:
> Right now, the main thing we need to fix is the read flow. I’m completely fine with either option 1 or 2. If you want, feel free to implement it, otherwise, I’ll handle it soon. After that, we can rebase your new resource and merge everything into devel

Thanks for the quick response. I will open a separate PR.

anton-sidelnikov:
@TobiWo, now we can gracefully deprecate this parameter and introduce your resource, #3375 merged

TobiWo:
> @TobiWo, now we can gracefully deprecate this parameter and introduce your resource, #3375 merged

Will apply the last steps tomorrow morning.

TobiWo:
Hi @anton-sidelnikov, 

* changes from both PRs are rebased into the feature branch
* tests run successfully
* PR description updated

PR is ready for final review.

anton-sidelnikov (approved):

Issue #3374, opened 2026-05-18
[VPC] `vpc_v1.secondary_cidr` Read claims out-of-band CIDRs into state


TobiWo:
## Terraform provider version

```bash
opentelekomcloud >= 1.36.61
```

## Affected Resource(s)

- opentelekomcloud_vpc_v1

## Terraform Configuration Files

```hcl
# Minimal repro: vpc_v1 without secondary_cidr in HCL.
resource "opentelekomcloud_vpc_v1" "test" {
  name = "tf-vpc-readbug"
  cidr = "192.168.0.0/16"
}
```

## Debug Output/Panic Output

No panic. The bug is visible from the source and from `terraform plan` after an external CIDR is added to the VPC's `extend_cidrs` (via the OTC console, another module, or the new `opentelekomcloud_vpc_secondary_cidr_v3` resource).

## Steps to Reproduce

1. `terraform apply` the configuration above.
2. Out of band, add a secondary CIDR to the same VPC's `extend_cidrs` list. For example via the OTC console, or by calling the v3 API: `PUT /v3/{project_id}/vpc/vpcs/{vpc_id}/add-extend-cidr`.
3. `terraform plan` against the unchanged HCL.

## Expected Behavior

`terraform plan` produces an empty diff. The user did not declare `secondary_cidr` in HCL, so Terraform should not claim ownership of any CIDR present on the cloud side.

## Actual Behavior

The plan proposes removing the externally-added CIDR:

```text
# opentelekomcloud_vpc_v1.test will be updated in-place
~ resource "opentelekomcloud_vpc_v1" "test" {
      id             = "..."
      name           = "tf-vpc-readbug"
    - secondary_cidr = "23.9.0.0/16" -> null
  }
```

If the user runs `terraform apply`, the `secondary_cidr` field is cleared from state and the cloud-side CIDR is removed via `RemoveSecondaryCidr`. Anything else managing that CIDR is overwritten.

## Important Factoids

- Root cause: `readSecondaryCidr` in `opentelekomcloud/services/vpc/resource_opentelekomcloud_vpc_v1.go:97-114` always calls `VpcV3.Get` and writes any non-empty `extend_cidrs[0]` into `secondary_cidr` state, even when the user did not set the field in HCL.
- This Read-side state pollution is independent of the destructive Update path that #3373 fixed. #3373 covered the `secondary_cidr = "x" → ""` removal flow on Update. The Read path was left untouched.
[...]

PR #3375, opened 2026-05-18
[Fix] VPC: vpc_v1 readSecondaryCidr leaks out-of-band CIDRs into state

TobiWo:

Fix a Read-side state pollution bug on `opentelekomcloud_vpc_v1`. `readSecondaryCidr` in `opentelekomcloud/services/vpc/resource_opentelekomcloud_vpc_v1.go` always hits the v3 API and writes any `extend_cidrs[0]` it finds into the `secondary_cidr` attribute, even when the user did not set the field in HCL. Any CIDR added to the same VPC by another tool (console, module, or `opentelekomcloud_vpc_secondary_cidr_v3`) is silently claimed by `vpc_v1` state. The next `terraform plan` proposes removing it, and `terraform apply` issues `RemoveSecondaryCidr` against the cloud, fighting the other tool.

The fix is an early-return guard at the top of `readSecondaryCidr`:

```go
if d.Get("secondary_cidr").(string) == "" {
    return nil
}
```

Users who set the field see no behaviour change. The v3 API is still consulted, state is still updated, and the existing Update path (fixed in #3373) still runs. Users who do not set the field keep `secondary_cidr` as `""` in state regardless of cloud-side `extend_cidrs`.

This unblocks PR #3371's `TestAccVpcSecondaryCidrV3_basic`, which was failing because of this Read-side pollution when the new resource adds a CIDR to a VPC managed by `vpc_v1`.

The change is intentionally minimal. It does not pre-empt the planned deprecation of `secondary_cidr` on `vpc_v1`, which will follow in a separate PR.

## PR Checklist

* [x] Refers to: #3374 
* [x] Tests added/passed.
* [ ] Documentation updated. (N/A — no schema or user-visible behaviour change for users who set the field; the fixed behaviour matches what the doc already implies)
* [ ] Schema updated. (N/A)
* [x] Release notes added.

## Acceptance Steps Performed

```text
=== RUN   TestAccVpcV1_basic
=== PAUSE TestAccVpcV1_basic
=== RUN   TestAccVpcV1_secondaryCidr
=== PAUSE TestAccVpcV1_secondaryCidr
=== RUN   TestAccVpcV1_readIgnoresExternalCidrs
=== PAUSE TestAccVpcV1_readIgnoresExternalCidrs
=== RUN   TestAccVpcV1_timeout
=== PAUSE TestAccVpcV1_timeout
=== CONT  TestAccVpcV1_basic
=== CONT  TestAccVpcV1_readIgnoresExternalCidrs
=== CONT  TestAccVpcV1_timeout
=== CONT  TestAccVpcV1_secondaryCidr
--- PASS: TestAccVpcV1_timeout (88.86s)
--- PASS: TestAccVpcV1_readIgnoresExternalCidrs (112.54s)
--- PASS: TestAccVpcV1_basic (156.84s)
--- PASS: TestAccVpcV1_secondaryCidr (182.50s)
PASS
```

The new `TestAccVpcV1_readIgnoresExternalCidrs` is a regression test for this bug. It applies a `vpc_v1` block without `secondary_cidr`, calls `VpcV3.AddSecondaryCidr` directly to inject an out-of-band CIDR, then runs `PlanOnly: true` and expects an empty plan. Without the fix the plan proposes removing the injected CIDR.

### DIFF
--- a/opentelekomcloud/acceptance/vpc/resource_opentelekomcloud_vpc_v1_test.go
+++ b/opentelekomcloud/acceptance/vpc/resource_opentelekomcloud_vpc_v1_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/opentelekomcloud/gophertelekomcloud/openstack/networking/v1/vpcs"
+	VpcV3 "github.com/opentelekomcloud/gophertelekomcloud/openstack/vpc/v3/vpcs"
 	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/common"
 	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/common/quotas"
 	"github.com/opentelekomcloud/terraform-provider-opentelekomcloud/opentelekomcloud/acceptance/env"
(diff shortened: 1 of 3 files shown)


TobiWo:
Hi @anton-sidelnikov , this is the discussed corresponding PR for #3371 .

anton-sidelnikov (approved):

anton-sidelnikov:
@TobiWo, okay, great, thank you!

Issue #3380, opened 2026-05-22
ECS: sync metadata to state in compute_instance_v2


HiraSiddiqui:
# Affected Resource(s)
- opentelekomcloud_compute_instance_v2

# Description
The metadata attribute on opentelekomcloud_compute_instance_v2 is never written back to Terraform state during a normal Read (refresh). After apply, the configured key-value pairs are sent to the OTC API and visible in the console, but on the next plan or refresh, metadata in state is empty. Only all_metadata (the full API response including system/provider-injected keys) is populated.

**The root cause is two issues combined:**

1. The metadata schema attribute is missing Computed: true, so any d.Set("metadata", ...) call in the Read function is silently discarded by the Terraform SDK.

2. The Read function never calls the dedicated /servers/{id}/metadata endpoint. It only sets all_metadata from the full server response.

**Scenarios where it breaks**

- **Drift detection broken:** A key is deleted out-of-band (via OTC console or API). On next plan, Terraform reports no changes instead of detecting the drift, because metadata in state is empty and there is nothing to compare against.

- **Import produces incomplete state:** After terraform import, metadata is empty in state even if the instance has user-defined keys. A subsequent plan shows changes that were never intended.

- **All_metadata is not a substitute:** Users who switch to reading all_metadata get system/provider-injected keys mixed in (image_name, os_type, etc.), making it unsuitable for programmatic use.

# Proposal

The /servers/{id}/metadata endpoint (Nova API) returns only user-managed keys, not system keys. This is the correct endpoint to use for syncing metadata state, as filtering server. 

Proposal is that after setting all_metadata, call the dedicated /servers/{id}/metadata endpoint (which returns only user-managed keys, not system keys) and write the result to metadata state. This will fix the diff if there was an out of band change.

# References
- Nova Metadata API: GET /v2/{project_id}/servers/{server_id}/metadata
https://docs.otc.t-systems.com/elastic-cloud-server/api-ref/native_openstack_nova_apis/metadata_management/querying_ecs_metadata.html#en-us-topic-0065817713

muneeb-jan:
Hi @HiraSiddiqui 

The changes should be available in the latest release v1.36.67. 
Closing the issue for now. Please reopen if the issue persists.

PR #3382, opened 2026-05-22
[Fix] ECS: sync metadata to state in compute_instance_v2

HiraSiddiqui:
Issue for this PR opened here: 
https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/issues/3380

### DIFF
--- a/opentelekomcloud/acceptance/ecs/resource_opentelekomcloud_compute_instance_v2_test.go
+++ b/opentelekomcloud/acceptance/ecs/resource_opentelekomcloud_compute_instance_v2_test.go
@@ -261,8 +261,13 @@ func TestAccComputeV2Instance_metadata(t *testing.T) {
 				Config: testAccComputeV2InstanceMetadata,
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckComputeV2InstanceExists(resourceInstanceV2Name, &instance),
+					// verify keys are present on the API-side server object
 					testAccCheckComputeV2InstanceMetadata(&instance, "foo", "bar"),
 					testAccCheckComputeV2InstanceMetadata(&instance, "abc", "def"),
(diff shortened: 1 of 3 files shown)


anton-sidelnikov:
Hi @HiraSiddiqui, this PR makes metadata track values read from the API, that means metadata added outside Terraform can be pulled into Terraform state and later deleted by Terraform if it is not present in the user config.

Openstack Upstream introduced all_metadata to avoid exactly this problem:

-   metadata is the Terraform-managed input.
-   all_metadata is the read-only output with everything returned by the OpenStack Compute API, including metadata added outside Terraform.

So the correct behaviour is: Terraform should update only keys from metadata, and users should read external/full metadata from all_metadata. Your change mixes those responsibilities and can cause accidental metadata drift or deletion

You could check this by adding any metadata value outside of terraform and then apply config

HiraSiddiqui:
> * including metadata added outside Terraform.

Hi @anton-sidelnikov,

Thanks for your comments. Let me try to explain my thought process on what I'm trying to fix.

The compute_instance_v2 has two properties "metadata" and "all_metadata". 

The all_metadata is the property that reads API on real time and returns all the parameters including the ones added/updated/deleted out of band. But the metadata property is actually the one that's used to detect changes in the config vs the deployed state 

`see d.hasChange("metadata") inside resourceComputeInstanceV2Update function`. 

However, the metadata property is never reconciled with the out-of-band changes. As a result, the metadata state will not be reconciled with the deployed state.

Therefore, I did the following two changes in my PR: 

1. Make metadata property updatable by adding the compute property
2. When the all_metadata is being read, also simultaneously read the changes into the metadata property as well. (this part can be cleaned up a bit in my PR but the metadata property needs to have the recent cloud state)

Let me know what's your reading on this.

anton-sidelnikov (approved):

Issue #3420, opened 2026-06-10
Add sso_type to opentelekomcloud_identity_provider_v3


HiraSiddiqui:
## Affected Package(s)

`opentelekomcloud_identity_provider_v3`

## Description

The `opentelekomcloud_identity_provider_v3` package does not support the `sso_type` field when creating or reading an identity provider. The OTC IAM Federation API accepts an `sso_type` field that determines how federated identities are mapped:

- `virtual_user_sso`: federated users are mapped to a single virtual user (the API default).
- `iam_user_sso`: federated users are mapped to existing IAM users.

Because `CreateOpts` has no field for it, every identity provider created through the SDK is created with the API's default (`virtual_user_sso`). And because the `Provider` result struct does not parse `sso_type`, callers cannot read the value back - even when a provider was created out-of-band with `iam_user_sso`.

This blocks the `terraform-provider-opentelekomcloud` resource `opentelekomcloud_identity_provider_v3` from exposing the attribute, since the provider can only send and read fields the SDK provides.

## Scenarios where this is a problem

- **`iam_user_sso` is unreachable via the SDK:** callers that federate an external IdP (e.g. Azure AD, ADFS) and want federated logins mapped to real IAM users cannot express this through `CreateOpts`.

- **Result struct drops the value:** even when an identity provider is created out-of-band with `iam_user_sso`, `providers.Get(...).Extract()` cannot surface the value because `Provider` has no `SSOType` field, so it is silently dropped from the parsed response.

- **Immutable, create-time only:** `sso_type` can only be set at creation and cannot be modified afterwards, so it must be supported on the `Create` path i.e. there is no post-hoc workaround.

## Proposal

Support `sso_type` in the identity provider package:

- Add `SSOType string` with tag `json:"sso_type,omitempty"` to `CreateOpts` (`requests.go`), so the field is sent on creation. Document it as create-only, immutable after creation, with allowed values `virtual_user_sso` (API default) and `iam_user_sso`.
- Add `SSOType string` with tag `json:"sso_type"` to the `Provider` result struct (`results.go`), so it is parsed from API responses (Create and Get).
- Update the test fixtures (`testing/fixtures.go`) so the Get/Create response fixture includes `sso_type`, and update the request tests (`testing/requests_test.go`) to assert the field on both `Create` and `Get`.

## Downstream

[...]

PR #3421, opened 2026-06-10
[Feat] Add sso_type to opentelekomcloud_identity_provider_v3

HiraSiddiqui:
The opentelekomcloud_identity_provider_v3 package does not support the sso_type field when creating or reading an identity provider. The OTC IAM Federation API accepts an sso_type field that determines how federated identities are mapped. However, the parameter was not available to be set in SDK and correspondingly in the terraform provider. 

This PR fixes it.

## PR Checklist

* [x] Refers to: https://github.com/opentelekomcloud/terraform-provider-opentelekomcloud/issues/3420
* [x] Tests added/passed.
* [x] Documentation updated.
* [x] Schema updated.
* [x] Release notes added.

## Acceptance Tests Performed
```
=== RUN   TestAccIdentityV3ProviderBasic
--- PASS: TestAccIdentityV3ProviderBasic (147.36s)

=== RUN   TestAccIdentityV3ProviderSSOType
--- PASS: TestAccIdentityV3ProviderSSOType (93.51s)

=== RUN   TestAccIdentityV3ProviderIAMUserSSO
--- PASS: TestAccIdentityV3ProviderIAMUserSSO (90.46s)
```

### DIFF
--- a/opentelekomcloud/services/iam/resource_opentelekomcloud_identity_provider_v3.go
+++ b/opentelekomcloud/services/iam/resource_opentelekomcloud_identity_provider_v3.go
@@ -72,11 +72,16 @@ func resourceIdentityProviderV3Create(ctx context.Context, d *schema.ResourceDat
 		return fmterr.Errorf(clientCreationFail, err)
 	}
 
+	ssoType := d.Get("sso_type").(string)
+	if ssoType == "" {
+		ssoType = ssoTypeVirtualUser
+	}
+
 	opts := providers.CreateOpts{
 		ID:          d.Get("name").(string),
 		Description: d.Get("description").(string),
 		Enabled:     d.Get("enabled").(bool),
-		SSOType:     d.Get("sso_type").(string),
+		SSOType:     ssoType,
 	}
 
 	p, err := providers.Create(client, opts).Extract()
--- a/opentelekomcloud/services/iam/resource_opentelekomcloud_identity_provider_v3.go
+++ b/opentelekomcloud/services/iam/resource_opentelekomcloud_identity_provider_v3.go
@@ -6,6 +6,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	golangsdk "github.com/opentelekomcloud/gophertelekomcloud"
 	"github.com/opentelekomcloud/gophertelekomcloud/openstack/identity/v3/federation/providers"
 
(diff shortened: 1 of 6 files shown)


muneeb-jan (changes requested):
Please fetch the latest changes from gopher (devel) and push the updated (clean) go.mod to the PR. 

```
go get github.com/opentelekomcloud/gophertelekomcloud@devel
go mod tidy
```

That way the changes in the gopher structs will be reflected in eco/check.

Also, remove the DEV SHIM

HiraSiddiqui:
> Please fetch the latest changes from gopher (devel) and push the updated (clean) go.mod to the PR.
> 
> ```
> go get github.com/opentelekomcloud/gophertelekomcloud@devel
> go mod tidy
> ```
> 
> That way the changes in the gopher structs will be reflected in eco/check.
> 
> Also, remove the DEV SHIM

@muneeb-jan the requested changes have been implemented.

opentelekomcloud/services/iam/resource_opentelekomcloud_identity_provider_v3.go
```diff
@@ -72,11 +72,16 @@ func resourceIdentityProviderV3Create(ctx context.Context, d *schema.ResourceDat
 		return fmterr.Errorf(clientCreationFail, err)
 	}
 
+	ssoType := d.Get("sso_type").(string)
+	if ssoType == "" {
+		ssoType = ssoTypeVirtualUser
+	}
```
muneeb-jan: you can use default in this case in the schema itself.

```
"sso_type": {
	Type:     schema.TypeString,
	Optional: true,
	Computed: true,
	ForceNew: true,
	ValidateFunc: validation.StringInSlice([]string{
		ssoTypeVirtualUser, ssoTypeIAMUser,
	}, false),
Default: ssoTypeVirtualUser
},
```

In that case you need to mention in the docs that it defaults to virtual_user_sso in tf as well.

However, omitempty is added to the `SSOType` field already. So we can just avoid the default since the API defaults to virtual_sso_user. If no value is passed to sso_type, it would just omit this field when sending the request.

muneeb-jan:
Also, please add the tests in the description.

opentelekomcloud/services/iam/resource_opentelekomcloud_identity_provider_v3.go
HiraSiddiqui: Ok, your approach is cleaner so I updated accordingly. Only caveat is that Computed and Default cannot be in the same schema so I removed Computed and kept Default.

HiraSiddiqui:
> Also, please add the tests in the description.

I have added the tests in the PR description. Do you mean in the documentation?

anton-sidelnikov (approved):

muneeb-jan (approved):

PR #3513, opened 2026-08-17
[Fix] VPC: Retry in-use error on secondary CIDR delete

### LINKED ISSUE #3512: [VPC] `opentelekomcloud_vpc_secondary_cidr_v3` delete fails with `cidr is used by subnets`
# Terraform provider version

```bash
opentelekomcloud >= 1.36.63 (all versions with this resource)
```

# Affected Resource(s)

- opentelekomcloud_vpc_secondary_cidr_v3

# Terraform Configuration Files

```hcl
resource "opentelekomcloud_vpc_v1" "vpc" {
  name = "sec-cidr-race"
  cidr = "192.168.0.0/16"
}

resource "opentelekomcloud_vpc_secondary_cidr_v3" "sec" {
  vpc_id = opentelekomcloud_vpc_v1.vpc.id
  cidrs  = ["23.9.0.0/16"]
}

resource "opentelekomcloud_vpc_subnet_v1" "back" {
  name       = "back"
  cidr       = "23.9.0.0/24"
  gateway_ip = "23.9.0.1"
  vpc_id     = opentelekomcloud_vpc_v1.vpc.id

  depends_on = [opentelekomcloud_vpc_secondary_cidr_v3.sec]
}
```

# Debug Output/Panic Output

```
Error: error removing secondary CIDRs from VPC <vpc-id>: Bad request with: [PUT
https://vpc.eu-de.otc.t-systems.com/v3/<project>/vpc/vpcs/<vpc-id>/remove-extend-cidr],
error message: {"error_msg":"cidr is used by subnets","error_code":"VPC.0602"}
```

# Steps to Reproduce

1. `terraform apply` the configuration above
2. `terraform destroy`

The failure is timing-dependent. It occurs when the subnet carved from the secondary CIDR has not finished releasing its ports by the time the secondary CIDR removal is issued. Re-running `terraform destroy` typically succeeds, because the subnet has finished releasing in the meantime.

# Expected Behavior

The destroy completes. `VPC.0602 cidr is used by subnets` is a transient condition while the subnet is released, and the provider should wait it out.

# Actual Behavior

`resourceVpcSecondaryCidrV3Delete` issues a single, unretried `RemoveSecondaryCidr` call. Only a 404 is tolerated; any other error, including the transient in-use error, aborts the delete immediately:

```go
if _, err := VpcV3.RemoveSecondaryCidr(client, d.Id(), VpcV3.CidrOpts{
    Vpc: &VpcV3.AddExtendCidrOption{ExtendCidrs: cidrs},
}); err != nil {
    if _, ok := err.(golangsdk.ErrDefault404); ok {
        return nil
    }
    return fmterr.Errorf("error removing secondary CIDRs from VPC %s: %w", d.Id(), err)
}
```

Every other delete in this package waits out transient states with a `resource.StateChangeConf` loop — for example `waitForVpcSubnetDelete` in `resource_opentelekomcloud_vpc_subnet_v1.go`. This one does not.

# Important Factoids

- The resource has no `Timeouts` block at all, so adding a retry loop also requires adding one. Matching `opentelekomcloud_vpc_subnet_v1` gives a 10 minute delete timeout.
- OTC returns `VPC.0602` as **HTTP 400**, not 409. `RemoveSecondaryCidr` declares `OkCodes: []int{200}`, so the SDK maps the status straight to `golangsdk.ErrDefault400`  — the `Bad request with: [PUT ...]` wording above is that type's message format. A retry must therefore match on the error code in the response body rather than on 400 alone, so that unrelated bad requests still fail fast.
- A related but independent defect makes this failure more likely to be observed: `opentelekomcloud_vpc_subnet_v1` treats HTTP 400 on the delete call as successful deletion and drops a live subnet from state. The orphaned subnet then keeps the secondary CIDR in use while Terraform believes the VPC has no subnets. Tracked separately in `tasks/fix-vpc-subnet-400/`. Retrying here makes that defect non-fatal for teardown, but does not fix it.

TobiWo:

`opentelekomcloud_vpc_secondary_cidr_v3` failed to delete with `error removing secondary CIDRs from VPC ...: cidr is used by subnets` (`VPC.0602`) when a subnet carved from that secondary CIDR was still being released.

The delete issued a single, unretried `RemoveSecondaryCidr` call and tolerated only a 404, so the transient in-use error aborted the destroy. Re-running `terraform destroy` normally succeeded, because the subnet finished releasing in the meantime.

* Wrap the removal in a `resource.StateChangeConf` loop that retries while the API returns `VPC.0602`, mirroring `waitForVpcSubnetDelete` in the same package. A 404 is still treated as already removed; any other error fails immediately.
* Add a `Timeouts` block with a 10 minute delete timeout — the resource previously had none — matching `opentelekomcloud_vpc_subnet_v1`.
* Add an acceptance test creating a VPC, a secondary CIDR, and a subnet carved from that CIDR in one apply. The destroy at test end is the assertion.

Note on the status code: OTC returns `VPC.0602` as **HTTP 400**, not 409 as the error text might suggest. The retry matches on the error code in the response body rather than on 400 alone, so unrelated bad requests still fail fast. A 409 is also treated as retryable defensively, but is not the code observed in practice.

### Known related defect, not fixed here

`opentelekomcloud_vpc_subnet_v1` treats HTTP 400 on the delete call as successful deletion and drops a live subnet from state, which makes this failure more likely to be hit. Reclassifying that 400 affects every subnet delete in the provider and its original motivation is undocumented, so it is left for a separate PR. The retry added here makes it non-fatal for teardown.

## PR Checklist

* [x] Refers to: #3512 
* [x] Tests added/passed.
* [x] Documentation updated. (N/A — no user-facing schema or behavior change)
* [x] Schema updated. (`Timeouts` block added)
* [x] Release notes added.

## Acceptance Steps Performed

```
=== RUN   TestAccVpcSecondaryCidrV3_basic
=== PAUSE TestAccVpcSecondaryCidrV3_basic
=== RUN   TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
=== PAUSE TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
=== CONT  TestAccVpcSecondaryCidrV3_basic
=== CONT  TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
--- PASS: TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr (87.81s)
--- PASS: TestAccVpcSecondaryCidrV3_basic (209.93s)
=== RUN   TestAccVpcSecondaryCidrV3_basic
=== PAUSE TestAccVpcSecondaryCidrV3_basic
=== RUN   TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
=== PAUSE TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
=== CONT  TestAccVpcSecondaryCidrV3_basic
=== CONT  TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
--- PASS: TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr (92.91s)
--- PASS: TestAccVpcSecondaryCidrV3_basic (206.64s)
=== RUN   TestAccVpcSecondaryCidrV3_basic
=== PAUSE TestAccVpcSecondaryCidrV3_basic
=== RUN   TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
=== PAUSE TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
=== CONT  TestAccVpcSecondaryCidrV3_basic
=== CONT  TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr
--- PASS: TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr (92.01s)
--- PASS: TestAccVpcSecondaryCidrV3_basic (209.12s)
PASS
```

Run with `-count=3`, since a single green run does not prove a race is fixed.

Closes #3512

### DIFF
--- a/opentelekomcloud/acceptance/vpc/resource_opentelekomcloud_vpc_secondary_cidr_v3_test.go
+++ b/opentelekomcloud/acceptance/vpc/resource_opentelekomcloud_vpc_secondary_cidr_v3_test.go
@@ -88,6 +88,34 @@ func TestAccVpcSecondaryCidrV3_basic(t *testing.T) {
 	})
 }
 
+func TestAccVpcSecondaryCidrV3_subnetOnSecondaryCidr(t *testing.T) {
+	t.Parallel()
+	quotas.BookOne(t, quotas.Router)
+	vpcName := tools.RandomString("tf-acc-sec-cidr-sub-", 5)
+
+	var vpc VpcV3.Vpc
+	rc := common.InitResourceCheck(resourceVpcSecondaryCidrV3Name, &vpc, getVpcSecondaryCidrFunc)
+
+	// The destroy is the assertion: removing the secondary CIDR while the subnet carved
+	// from it is still draining returns 409, which the delete must wait out.
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { common.TestAccPreCheck(t) },
+		ProviderFactories: common.TestAccProviderFactories,
+		CheckDestroy:      rc.CheckResourceDestroy(),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccVpcSecondaryCidrV3WithSubnet(vpcName),
+				Check: resource.ComposeTestCheckFunc(
+					rc.CheckResourceExists(),
+					resource.TestCheckResourceAttr(resourceVpcSecondaryCidrV3Name, "cidrs.#", "1"),
+					resource.TestCheckResourceAttr(
+						"opentelekomcloud_vpc_subnet_v1.test", "cidr", "23.9.0.0/24"),
+				),
+			},
+		},
+	})
+}
+
 func testAccVpcSecondaryCidrV3OneCidr(vpcName string) string {
 	return fmt.Sprintf(`
 resource "opentelekomcloud_vpc_v1" "test" {
@@ -100,6 +128,29 @@ resource "opentelekomcloud_vpc_secondary_cidr_v3" "test" {
   cidrs  = ["23.9.0.0/16"]
 }
 `, vpcName)
+}
+
+func testAccVpcSecondaryCidrV3WithSubnet(vpcName string) string {
+	return fmt.Sprintf(`
+resource "opentelekomcloud_vpc_v1" "test" {
+  name = "%s"
+  cidr = "192.168.0.0/16"
+}
+
+resource "opentelekomcloud_vpc_secondary_cidr_v3" "test" {
+  vpc_id = opentelekomcloud_vpc_v1.test.id
+  cidrs  = ["23.9.0.0/16"]
+}
+
+resource "opentelekomcloud_vpc_subnet_v1" "test" {
+  name       = "%s-sub"
+  cidr       = "23.9.0.0/24"
+  gateway_ip = "23.9.0.1"
+  vpc_id     = opentelekomcloud_vpc_v1.test.id
+
+  depends_on = [opentelekomcloud_vpc_secondary_cidr_v3.test]
+}
+`, vpcName, vpcName)
 }
 
 func testAccVpcSecondaryCidrV3ThreeCidrs(vpcName string) string {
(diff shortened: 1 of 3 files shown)


anton-sidelnikov (approved):

anton-sidelnikov:
regate

